PR #6, opened 2015-05-12
Events

bitpshr:
This pull request includes `on`, `aspect`, `Evented`, handle functions in `lang`, and associated test updates.


package.json
```diff
@@ -1,29 +1,32 @@
 {
```
bryanforbes, line 1: This file should have hard tabs and not spaces

package.json
```diff
@@ -1,29 +1,32 @@
...
+    "grunt-text-replace": "0.4.0",
+    "grunt-ts": "3.0.0",
+    "grunt-tslint": "2.0.0",
+    "tslint": "2.1.1",
+    "intern": "2.2.2",
+    "requirejs": "2.1.17",
+    "typescript": "Arnavion/typescript-github#2015-05-07-e1d05150f6acef05452772192a4abdfc7eb671eb"
+  },
+  "dependencies": {
+    "sinon": "~1.14.1"
```
bryanforbes: This should be in `devDependencies` and also use a hard version instead of a fuzzy one

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
+import { Handle, EventObject } from './interfaces';
+import { createCompositeHandle } from './lang';
+import * as aspect from './aspect';
```
bryanforbes: Why not just import `on` from `aspect`?

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default class Evented {
+	emit(data: EventObject): boolean {
+		var type = '__on' + data.type;
+		var method: Function = (<any> this)[type];
+		if (method) {
+			return method.call(this, data);
+		}
+	};
```
bryanforbes: No need for a semi-colon here

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
...
+		var name = '__on' + type;
+		if (!(<any> this)[name]) {
+			Object.defineProperty(this, name, {
+				configurable: true,
+				value: undefined,
+				writable: true
+			});
+		}
+		return aspect.on(this, '__on' + type, listener);
+	};
```
bryanforbes: See previous note

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
+import { Handle, EventObject } from './interfaces';
+import { createCompositeHandle } from './lang';
+import * as aspect from './aspect';
+
+export default class Evented {
+	emit(data: EventObject): boolean {
+		var type = '__on' + data.type;
```
bryanforbes: `var` -> `const`

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
...
+		var method: Function = (<any> this)[type];
```
bryanforbes: `var` -> `const`

src/Evented.ts
```diff
@@ -0,0 +1,25 @@
...
+		if (method) {
+			return method.call(this, data);
+		}
+	};
+
+	on(type: string, listener: (event: EventObject) => void): Handle {
+		var name = '__on' + type;
```
bryanforbes: `var` -> `const`

src/lang.ts
```diff
@@ -1,6 +1,7 @@
 import { PropertyEvent, Observer } from './observers/interfaces';
 import * as ObjectObserver from './observers/ObjectObserver';
 import has from './has';
+import { Handle, EventObject } from './interfaces';
```
bryanforbes: Why is `EventObject` being imported?

src/lang.ts
```diff
@@ -213,3 +214,20 @@ export function partial(targetFunction: (...args: any[]) => any, ...suppliedArgs
...
+		destroy: function () {
+			this.destroy = function () {};
+			destructor.call(this);
+		}
+	};
+}
+
+export function createCompositeHandle(...handles: Handle[]): Handle {
+	return createHandle(function () {
+		for (var i = 0, handle: Handle; (handle = handles[i]); ++i) {
```
bryanforbes: Use a for-of loop

src/on.ts
```diff
@@ -0,0 +1,93 @@
+import { Handle, EventObject } from './interfaces';
+import { createHandle, createCompositeHandle } from './lang';
+import Evented from './Evented';
+
+// Only used for Evented and EventEmitter, as EventTarget uses EventListener
+type EventCallback = (event: {}) => void;
+
+interface ExtensionEvent {
```
bryanforbes: This will need to be exported since `on()` is using it

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+	(target: any, listener: EventCallback): Handle;
+}
+
+interface EventTarget {
```
bryanforbes: Needs to be exported

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+	addEventListener(event: string, listener: EventListener, capture?: boolean): Handle;
+	removeEventListener(event: string, listener: EventListener, capture?: boolean): void;
+}
+
+interface EventEmitter {
```
bryanforbes: Needs to be exported

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+    on(event: string, listener: EventCallback): EventEmitter;
```
bryanforbes: Tabs, not spaces

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+	removeListener(event: string, listener: EventCallback): EventEmitter;
+}
+
+export default function on(target: EventTarget, type: string, listener: EventListener, capture?: boolean): Handle;
+export default function on(target: EventTarget, type: ExtensionEvent, listener: EventListener, capture?: boolean): Handle;
+export default function on(target: EventTarget, type: [string|ExtensionEvent], listener: EventListener, capture?: boolean): Handle;
```
bryanforbes: What is `[string|ExtensionEvent]`?

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export default function on(target: EventEmitter, type: string, listener: EventCallback): Handle;
+export default function on(target: EventEmitter, type: ExtensionEvent, listener: EventCallback): Handle;
+export default function on(target: EventEmitter, type: [string|ExtensionEvent], listener: EventCallback): Handle;
+
+export default function on(target: Evented, type: string, listener: EventCallback): Handle;
+export default function on(target: Evented, type: ExtensionEvent, listener: EventCallback): Handle;
+export default function on(target: Evented, type: [string|ExtensionEvent], listener: EventCallback): Handle;
+
+export default function on(target: any, type: any, listener: EventListener, capture?: boolean): Handle {
```
bryanforbes: `listener` should be `any`

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+	if (type.call) {
+		return type.call(this, target, listener, capture);
+	}
+
+	if (!!type.shift) {
```
bryanforbes: This is a bad inference. This should probably be `Array.isArray(type)`.

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+		var handles: Handle[] = type.map(function (type: string): Handle {
+			return on(target, type, listener, capture);
+		});
+
+		return createCompositeHandle.apply(null, handles);
+	}
+
+	if (target.addEventListener && target.removeEventListener) {
+		target.addEventListener(type, listener, false);
```
bryanforbes: Two things:
1. The third argument needs to be `capture`.
2. If I pass the same listener into this function twice, `addEventListener` will not register it a second time. The listener should probably be wrapped with a function to make it unique.

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+		return createHandle(function () { target.removeEventListener(type, listener, false); });
+	}
+	else if (target.on && target.removeListener) {
+		target.on(type, listener);
```
bryanforbes: The same consideration in the previous note's second point needs to be considered here.

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+		return createHandle(function () { target.removeListener(type, listener); });
+	}
+	else if (target.on) {
+		return target.on(type, listener);
+	}
+	else {
+		throw new TypeError('Unknown event emitter object');
+	}
+};
```
bryanforbes: No semi-colon needed

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// TODO: What should this return type signature be? Boolean for cancel? The event object itself?
+export function emit(target: EventTarget, event: EventObject): boolean;
+export function emit(target: EventEmitter, event: EventObject): boolean;
+export function emit(target: Evented, event: EventObject): boolean;
+export function emit(target: any, event: any): boolean {
+	if (target.emit && target.removeListener) {
+		return target.emit(event.type, event);
+	}
+	else if (target.emit) {
```
bryanforbes: Shouldn't this be `target.emit && target.on`?

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+		return target.emit(event, event);
```
bryanforbes: If this is supporting `Evented`, there's no need to pass a second argument

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if (typeof target.emit === 'function' && !target.nodeType) {
```
bryanforbes: I'm not sure what this is trying to support

src/on.ts
```diff
@@ -0,0 +1,93 @@
...
+		return target.emit(event.type, event);
+	}
+
+	if (target.dispatchEvent && target.ownerDocument && target.ownerDocument.createEvent) {
+		var nativeEvent = target.ownerDocument.createEvent('HTMLEvents');
+		nativeEvent.initEvent(event.type, Boolean(event.bubbles), Boolean(event.cancelable));
+
+		for (var key in event) {
+			if (!(key in nativeEvent)) {
+				nativeEvent[key];
```
bryanforbes: I don't understand this...

tests/unit/on/browserOnly.ts
```diff
@@ -0,0 +1,21 @@
+import common from './common';
+import registerSuite = require('intern!object');
+import assert = require('intern/chai!assert');
+import on from 'src/on';
+import Evented from 'src/Evented';
+
+registerSuite({
+	name: 'on',
+
+	'DOM events': common({
```
bryanforbes: These tests don't actually test `on()`

tests/unit/on/common.ts
```diff
@@ -0,0 +1,142 @@
...
+			emittedEvent = { value: 'foo', type: testEventName };
+			emit(target, emittedEvent);
+			assert.strictEqual(listenerCallCount, 1);
+
+			emittedEvent = { value: 'bar', type: testEventName };
+			emit(target, emittedEvent);
+			assert.strictEqual(listenerCallCount, 2);
+		},
+
+		'.emit return value'() {
```
bryanforbes: Is there a reason this is commented out?

tests/unit/on/nodeOnly.ts
```diff
@@ -0,0 +1,18 @@
...
+import registerSuite = require('intern!object');
+import assert = require('intern/chai!assert');
+import on from 'src/on';
+import Evented from 'src/Evented';
+import {EventEmitter} from 'events';
+
+registerSuite({
+	name: 'on',
+
+	'node events': common({
```
bryanforbes: These tests aren't actually testing anything

typings/sinon/sinon.d.ts
```diff
@@ -0,0 +1,420 @@
+// Type definitions for Sinon 1.8.1
```
bryanforbes: This file should go into `tests/typings/sinon`

typings/tsd.d.ts
```diff
@@ -1 +1,2 @@
 /// <reference path="node/node.d.ts" />
+/// <reference path="sinon/sinon.d.ts" />
```
bryanforbes: This reference comment should go into `tests/typings/tsd.d.ts`

bryanforbes:
A couple of general comments:
1. There are several places where code doesn't conform to the style guide. That will need to be cleaned up.
2. The tests don't run in the browser because Sinon is not registered in the loader

tests/intern.ts
```diff
@@ -44,7 +44,8 @@ export var loader = {
 	// Packages that should be registered with the loader in each testing environment
 	packages: [
 		{ name: 'src', location: '_build/src' },
-		{ name: 'tests', location: '_build/tests' }
+		{ name: 'tests', location: '_build/tests' },
+		{ name: 'dojo', location: 'node_modules/intern/node_modules/dojo'}
```
bryanforbes: Sinon needs to be added here so tests run in the browser

tests/unit/lang.ts
```diff
@@ -301,5 +301,32 @@ registerSuite({
...
+		var count = 0;
+		function destructor(): void {
+			count++;
+		}
+		var handle = lang.createCompositeHandle(
+			lang.createHandle(destructor),
+			lang.createHandle(destructor)
+		);
+
+		handle.destroy();
```
bryanforbes, line 329: There should be another test that makes sure calling `destroy()` twice won't call the destructor two more times.

bryanforbes:
I thought of another thing: `tsconfig.json` needs to be updated. If you run `grunt dev`, it should do that for you.

src/lang.ts
bitpshr: I did not think `for...of` was supported by IE. Should we go with use `Array.prototype.forEach` for the same code cleanliness instead?

bryanforbes: `for...of` compiles to a `for` loop when targeting ES3 and ES5 ([reference](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#forof-support))

bitpshr: Nice. Thanks!

src/on.ts
bitpshr: This adds support for attaching a single handler for multiple events in one statement. Dojo 1.x used comma-separated string syntax and parsed it accordingly, adding unnecessary complexity. @kriszyp suggested supporting arrays of events in chat in passing. It's easy enough to remove if we want to nix it.

bryanforbes: I'm OK with that, however that type statement is not doing what you think it is. `[string|ExtensionEvent]` is an array of one `string` or `ExtenstionEvent`. You probably want `(string | ExtensionEvent)[]`.

bitpshr: Noted. Thanks for this clarification!

src/on.ts
bitpshr: It would help if I completed the right-hand side of this statement :)

src/on.ts
```diff
@@ -0,0 +1,97 @@
+import { Handle, EventObject } from './interfaces';
+import { createHandle, createCompositeHandle } from './lang';
+import Evented from './Evented';
+
+// Only used for Evented and EventEmitter, as EventTarget uses EventListener
+type EventCallback = (event: {}) => void;
```
bryanforbes, line 6: This also needs to be exported. Without this exported, dts-generator fails to generate a definition for the `on` module. You can check this with `grunt dist`.

src/on.ts
```diff
@@ -0,0 +1,97 @@
...
+			if (!(key in nativeEvent)) {
+				nativeEvent[key] = event[key];
+			}
+		}
+
+		return target.dispatchEvent(nativeEvent);
+	}
+
+	if (target.emit && target.removeListener) {
+		return target.emit(event.type, event);
```
bryanforbes, line 42: Shouldn't this return false since these events won't bubble?

src/on.ts
```diff
@@ -0,0 +1,97 @@
...
+	if (type.call) {
+		return type.call(this, target, listener, capture);
+	}
+
+	if (Array.isArray(type)) {
+		let handles: Handle[] = type.map(function (type: string): Handle {
+			return on(target, type, listener, capture);
+		});
+
+		return createCompositeHandle.apply(null, handles);
```
bryanforbes, line 75: This can be changed to `createCompositeHandle(...handles);`

src/on.ts
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	const callback = function () {
+		listener.apply(this, arguments);
+	}
+
+	if (target.addEventListener && target.removeEventListener) {
+		target.addEventListener(type, callback, capture);
+		return createHandle(function () { target.removeEventListener(type, callback, capture); });
```
bryanforbes, line 84: This shouldn't be one line

src/on.ts
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if (target.on && target.removeListener) {
+		target.on(type, callback);
+		return createHandle(function () { target.removeListener(type, callback); });
```
bryanforbes, line 89: This also shouldn't be one line

src/on.ts
bitpshr, line 6: Noted. For future reference, how did you discover this? Manually checking the output?

src/on.ts
bitpshr, line 42: This depends on our decision to normalize return values of `emit`. Do we pave over `EventEmitter.emit`'s functionality that returns `true` or `false` depending on if callbacks were registered? For now I left this in, but calling `target.emit(event.type, event);` then explicitly returning `false` would be fine if that's preferable.

src/on.ts
bryanforbes, line 6: Yeah. Eventually dts-generator will emit an error.

bryanforbes:
I went ahead, cleaned this up, and merged it in c5f70a672d78d3bf7a6d99481fa1e801469f22db.

PR #7, opened 2015-05-12
reflect

bitpshr:
This PR includes an implementation of the `reflect` proposal. No tests have been written yet.


dylans:
This is unfortunately out of date... if redone, should live in shim, have tests, and should verify it's up to date with the current spec (e.g. Reflect. enumerate was removed from ES8).

PR #8, opened 2015-05-12
Map

bitpshr:
This PR includes an implementation of the `Map` class and associated tests.

### DIFF
--- a/src/Map.ts
+++ b/src/Map.ts
@@ -1,143 +1,171 @@
-import {ArrayLike} from './array';
+import { hasClass } from './decorators';
 import {is} from './object';
+import global from './global';
 
-/**
- * An implementation of the ES2015 Map specification using arrays
- * instead of iterators. The iterator protocol is not implemented.
- */
-export class Map<K, V> {
-	private mapKeys: K[] = [];
-	private mapValues: V[] = [];
+module Native {
+	interface NativeMap<K, V> {
+		size: number;
+		clear(): void;
+		delete(key: K): boolean;
+		entries(): any;
+		forEach(callback: (key: K, value: V, instance: Map<K, V>) => any, context?: {}): void;
+		get(key: K): V;
+		has(key: K): boolean;
+		keys(): K[];
+		set(key: K, value: V): NativeMap<K, V>;
+		values(): V[];
+	}
 
-	/*
-	 * An alternative to Array.prototype.indexOf using Object.is
-	 * to check for equality. See http://mzl.la/1zuKO2V
+	// Since ES6 classes aren't inheritable, create a thin wrapper around
+	// the native WeakMap with an inheritable TS class
+	export class Map<K, V> {
+		private _map: NativeMap<K, V>;
+
+		constructor(iterable?: any) {
+			Object.defineProperty(this, '_map', {
+				value: new global.Map(iterable)
+			});
+		}
+
+		get size(): number {
+			return this._map.size;
+		}
+
+		clear(): void {
+			return this._map.clear();
+		}
+
+		delete(key: K): boolean {
+			return this._map.delete(key);
+		}
+
+		entries(): any[] {
+			return this._map.entries();
+		}
+
+		forEach(callback: (key: K, value: V, instance: Map<K, V>) => any, context?: {}): void {
+			return this._map.forEach(callback, context);
+		}
+
+		get(key: K): V {
+			return this._map.get(key);
+		}
+
+		has(key: K): boolean {
+			return this._map.has(key);
+		}
+
+		keys(): K[] {
+			return this._map.keys();
+		}
+
+		set(key: K, value: V): Native.Map<K, V> {
+			this._map.set(key, value);
+			return this;
+		}
+
+		values(): V[] {
+			return this._map.values();
+		}
+	}
+}
+
+module Shim {
+	/**
+	 * An shimmed implementation of the ES2015 Map specification using arrays
+	 * instead of iterators. The iterator protocol is not implemented.
 	 */
-	private indexOfKey(keys: K[], key: K): number {
-		for (var i = 0; i < keys.length; i++) {
-			if (is(keys[i], key)) {
-				return i;
+	export class Map<K, V> {
+		private _keys: K[] = [];
+		private _values: V[] = [];
+
+		/*
+		 * An alternative to Array.prototype.indexOf using Object.is
+		 * to check for equality. See http://mzl.la/1zuKO2V
+		 */
+		private _indexOfKey(keys: K[], key: K): number {
+			for (var i = 0; i < keys.length; i++) {
+				if (is(keys[i], key)) {
+					return i;
+				}
+			}
+			return -1;
+		}
+
+		constructor(iterable?: any) {
+			if (!(this instanceof Map)) {
+				throw new TypeError('Constructor Map requires "new"');
+			}
+
+			if (iterable) {
+				for (let [key, value] of iterable) {
+					this.set(key, value);
+				}
 			}
 		}
-		return -1;
-	}
 
-	/**
-	 * Returns the value associated with a given key.
-	 *
-	 * @param key The key to look up
-	 * @return the value if one exists or undefined
-	 */
-	get(key: K): V {
-		var index = this.indexOfKey(this.mapKeys, key);
-		return index < 0 ? undefined : this.mapValues[index];
-	}
-
-	/**
-	 * Sets the value associated with a given key.
-	 *
-	 * @param key The key to define a value to
-	 * @param value The value to assign
-	 * @return A Map instance
-	 */
-	set(key: K, value: V): Map<K, V> {
-		var index = this.indexOfKey(this.mapKeys, key);
-		index = index < 0 ? this.mapKeys.length : index;
-		this.mapKeys[index] = key;
-		this.mapValues[index] = value;
-		return this;
-	}
-
-	/**
-	 * Deletes a given key and its associated value.
-	 *
-	 * @param key The key to delete
-	 * @return true if the key exists, false if it does not
-	 */
-	delete(key: K): boolean {
-		var index = this.indexOfKey(this.mapKeys, key);
-		if (index < 0) {
-			return false;
+		get size(): number {
+			return this._keys.length;
 		}
-		this.mapKeys.splice(index, 1);
-		this.mapValues.splice(index, 1);
-		return true;
-	}
 
-	/**
-	 * Checks for the presence of a given key.
-	 *
-	 * @param key The key to check for
-	 * @return true if the key exists, false if it does not
-	 */
-	has(key: K): boolean {
-		return this.indexOfKey(this.mapKeys, key) > -1;
-	}
+		clear(): void {
+			this._keys.length = this._values.length = 0;
+		}
 
-	/**
-	 * Deletes all keys and their associated values.
-	 */
-	clear(): void {
-		this.mapKeys.length = this.mapValues.length = 0;
-	}
+		delete(key: K): boolean {
+			var index = this._indexOfKey(this._keys, key);
+			if (index < 0) {
+				return false;
+			}
+			this._keys.splice(index, 1);
+			this._values.splice(index, 1);
+			return true;
+		}
 
-	/**
-	 * Returns an array of map keys in order of insertion.
-	 *
-	 * @return an array of keys in order of insertion
-	 */
-	keys(): K[] {
-		return this.mapKeys;
-	}
+		entries(): any[] {
+			var entries: any[] = [];
+			this._keys.forEach(function (key: K) {
+				entries.push([key, this.get(key)]);
+			}, this);
+			return entries;
+		}
 
-	/**
-	 * Returns the number of key / value pairs in the Map
-	 *
-	 * @return the number of key / value pairs in the Map
-	 */
-	get size(): number {
-		return this.mapKeys.length;
-	}
+		forEach(callback: (key: K, value: V, mapInstance: Map<K, V>) => any, context?: {}) {
+			// don't use this.entries to avoid second forEach call
+			this._keys.forEach(function (key, index) {
+				callback.call(context, key, this.get(key), this);
+			}, this);
+		}
 
-	/**
-	 * Returns an array of map values in order of insertion.
-	 *
-	 * @return an array of values in order of insertion
-	 */
-	values(): V[] {
-		return this.mapValues;
-	}
+		get(key: K): V {
+			var index = this._indexOfKey(this._keys, key);
+			return index < 0 ? undefined : this._values[index];
+		}
 
-	/**
-	 * Returns an array of two-value tuples in the form
-	 * of [key, value] in order of insertion.
-	 *
-	 * @return an array of entries in order of insertion
-	 */
-	// TODO: change this any type to K, V tuple
-	entries(): any {
-		var entries: any[] = [];
-		this.mapKeys.forEach(function (key: K) {
-			entries.push([key, this.get(key)]);
-		}, this);
-		return entries;
-	}
+		has(key: K): boolean {
+			return this._indexOfKey(this._keys, key) > -1;
+		}
 
-	/**
-	 * Executes a given function for each map entry. The function 
-	 * is invoked with three arguments: the element value, the
-	 * element key, and the associated Map instance.
-	 *
-	 * @param callback The function to execute for each map entry,
-	 * @param context The value to use for `this` for each execution
-	 * of the calbackv
-	 */
-	forEach(callback: (key: K, value: V, mapInstance: Map<K, V>) => any, context?: {}) {
-		// don't use this.entries to avoid second forEach call
-		this.mapKeys.forEach(function (key, index) {
-			callback.call(context, key, this.get(key), this);
-		}, this);
+		keys(): K[] {
+			return this._keys;
+		}
+
+		set(key: K, value: V): Shim.Map<K, V> {
+			var index = this._indexOfKey(this._keys, key);
+			index = index < 0 ? this._keys.length : index;
+			this._keys[index] = key;
+			this._values[index] = value;
+			return this;
+		}
+
+		values(): V[] {
+			return this._values;
+		}
 	}
+}
+
+@hasClass('map', Native.Map, Shim.Map)
+export default class Map<K, V> {
 
 	/**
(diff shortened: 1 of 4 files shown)


src/Map.ts
```diff
@@ -0,0 +1,261 @@
+import { hasClass } from './decorators';
+import {is} from './object';
```
bryanforbes: Space within braces

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+module Native {
+	interface NativeMap<K, V> {
+		size: number;
+		clear(): void;
+		delete(key: K): boolean;
+		entries(): any;
+		forEach(callback: (key: K, value: V, instance: Map<K, V>) => any, context?: {}): void;
+		get(key: K): V;
+		has(key: K): boolean;
+		keys(): K[];
```
bryanforbes: The native `Map#keys()` actually returns an iterator

src/Map.ts
```diff
@@ -0,0 +1,261 @@
+import { hasClass } from './decorators';
+import {is} from './object';
+import global from './global';
+
+module Native {
+	interface NativeMap<K, V> {
+		size: number;
+		clear(): void;
+		delete(key: K): boolean;
+		entries(): any;
```
bryanforbes: The native `Map#entries()` actually returns an iterator and the items are of type `[K, V]`.

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+		forEach(callback: (key: K, value: V, instance: Map<K, V>) => any, context?: {}): void;
+		get(key: K): V;
+		has(key: K): boolean;
+		keys(): K[];
+		set(key: K, value: V): NativeMap<K, V>;
+		values(): V[];
```
bryanforbes: The native `Map#values()` returns an iterator.

src/Map.ts
kfranqueiro: Just refreshing the earlier line note here, that `{is}` should be `{ is }` here.

src/Map.ts
```diff
@@ -0,0 +1,261 @@
+import { hasClass } from './decorators';
+import {is} from './object';
+import global from './global';
+
+module Native {
```
kfranqueiro: At this point, Forbes and I are thinking that it may be more prudent for core's `Map` module to settle for implementing a "Map-like" interface (what the `Shim` currently does) rather than trying to wrap native if it exists.  In particular, this will avoid conundrums around variance in platforms' support of Iterable objects.

So we should be able to remove `module Native` entirely.

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+		get size(): number {
+			return this._keys.length;
+		}
+
+		clear(): void {
+			this._keys.length = this._values.length = 0;
+		}
+
+		delete(key: K): boolean {
+			var index = this._indexOfKey(this._keys, key);
```
kfranqueiro: Replace `var` with `const` here

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+			}
+			this._keys.splice(index, 1);
+			this._values.splice(index, 1);
+			return true;
+		}
+
+		entries(): any[] {
+			var entries: any[] = [];
+			this._keys.forEach(function (key: K) {
+				entries.push([key, this.get(key)]);
```
kfranqueiro: It should be possible to make this more efficient, if you include the 2nd (index) parameter in the `forEach` callback and then perform the index lookup in `_values` directly.

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+			var index = this._indexOfKey(this._keys, key);
+			return index < 0 ? undefined : this._values[index];
+		}
+
+		has(key: K): boolean {
+			return this._indexOfKey(this._keys, key) > -1;
+		}
+
+		keys(): K[] {
+			return this._keys;
```
kfranqueiro: Consider slicing this to avoid breaking the map if someone mutates the array it returns.  (Might be worth adding a unit test for.)

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+		set(key: K, value: V): Shim.Map<K, V> {
+			var index = this._indexOfKey(this._keys, key);
+			index = index < 0 ? this._keys.length : index;
+			this._keys[index] = key;
+			this._values[index] = value;
+			return this;
+		}
+
+		values(): V[] {
+			return this._values;
```
kfranqueiro: Consider slicing this to avoid breaking the map if someone mutates the array it returns.

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+		}
+	}
+}
+
+@hasClass('map', Native.Map, Shim.Map)
+export default class Map<K, V> {
```
kfranqueiro: This will no longer be needed when we simplify to always returning a basic implementation regardless of native support.  (Just take the exported class out of the internal Shim module above.)

src/Map.ts
```diff
@@ -0,0 +1,261 @@
...
+		set(key: K, value: V): NativeMap<K, V>;
+		values(): V[];
+	}
+
+	// Since ES6 classes aren't inheritable, create a thin wrapper around
+	// the native WeakMap with an inheritable TS class
+	export class Map<K, V> {
+		private _map: NativeMap<K, V>;
+
+		constructor(iterable?: any) {
```
kfranqueiro: I think this should be able to use `array.ArrayLike` (from `src/array.ts`) instead of `any`.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			assert.doesNotThrow(function () {
+				map = new Map<number, string>([]);
+			});
+		},
+
+		'array like data'() {
+			assert.doesNotThrow(function () {
+				map = new Map<number, string>({
+					length: 1,
+					0: [1337, 'bar']
```
kfranqueiro: Missing spaces inside brackets.  I'd also settle for a lower number ;)

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+		'object key'() {
+			map = new Map<{}, string>();
+			var object = Object.create(null);
+			map.set(object, 'abc');
+			assert.strictEqual(map.get(object), 'abc');
+		},
+
+		'array key'() {
+			map = new Map<any[], string>();
+			var array = new Array();
```
kfranqueiro: Why not `[]` instead of `new Array()`?

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+		},
+
+		'returns instance'() {
+			map = new Map<string, string>();
+			assert.instanceOf(map.set('foo', 'bar'), Map);
+			assert.strictEqual(map.set('foo', 'bar'), map);
+		},
+
+		'key exists'() {
+			map = new Map<number, string>([[1337, 'abc']]);
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			map.set(1337, 'def');
+			assert.strictEqual(map.get(1337), 'def');
+		}
+	},
+
+	get: {
+		before() {
+			map = new Map<number, string>([
+				[0, 'a'],
```
kfranqueiro, line 97: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+		assert.deepEqual(map.keys(), [
+			0, '1', object, array, foo
+		]);
+	},
+
+	values() {
+		function foo() {}
+		var object = Object.create(null);
+		var array = new Array();
+		map = new Map<any, any>([
```
kfranqueiro: Given the example below, this could be `<number, any>`, right?

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			[4, foo],
+			[5, undefined]
+		]);
+		assert.deepEqual(map.values(), [
+			0, '1', object, array, foo, undefined
+		]);
+	},
+
+	entries: {
+		'empty map'() {
```
kfranqueiro, line 204: Is there a reason why we test this here but not for keys/values?

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			map = new Map<void, void>();
+			assert.doesNotThrow(function () {
+				map.clear();
+			});
+		},
+
+		'nonempty map'() {
+			map = new Map<number, string>([
+				[1337, 'abc']
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+
+		'key not found'() {
+			assert.isFalse(map.delete('foo'));
+		}
+	},
+
+	has: {
+		before() {
+			map = new Map<number, string>([
+				[1337, 'abc']
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+
+		'key not found'() {
+			assert.isUndefined(map.get(1337));
+		}
+	},
+
+	'delete': {
+		before() {
+			map = new Map<number, string>([
+				[1337, 'abc']
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			assert.strictEqual(map.size, 0);
+		}
+	},
+
+	keys() {
+		function foo() {}
+		var object = Object.create(null);
+		var array = new Array();
+		map = new Map<any, any>([
+			[0, 0],
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			0, '1', object, array, foo
+		]);
+	},
+
+	values() {
+		function foo() {}
+		var object = Object.create(null);
+		var array = new Array();
+		map = new Map<any, any>([
+			[0, 0],
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+
+	entries: {
+		'empty map'() {
+			map = new Map<void, void>();
+			assert.deepEqual(map.entries(), []);
+		},
+
+		'nonempty map'() {
+			map = new Map<number, string>([
+				[1, 'a'],
```
kfranqueiro: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+		}
+	},
+
+	iteration: {
+		before() {
+			function foo() {}
+			var object = Object.create(null);
+			var array = new Array();
+			map = new Map<any, any>([
+				[0, 0],
```
kfranqueiro, line 229: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+			assert.doesNotThrow(function () {
+				map = new Map<number, string>({
+					length: 1,
+					0: [1337, 'bar']
+				});
+			});
+		}
+	},
+
+	set: {
```
kfranqueiro, line 40: We should probably test that `size` updates, either in one of these tests, or in a separate test after the `set` tests.

tests/unit/Map.ts
```diff
@@ -0,0 +1,239 @@
...
+				assert.strictEqual(map, mapInstance);
+			})
+		},
+
+		'times executed'() {
+			var counter = 0;
+			map.forEach(function (key, value, mapInstance) {
+				counter++;
+			});
+			assert.strictEqual(counter, 6);
```
kfranqueiro: If you actually store the array you're passing to `new Map`, you could reuse its `length` here.

tests/unit/Map.ts
bitpshr, line 40: Done. Not sure why this isn't updating.

bitpshr:
This should be ready for another review. Thanks @kfranqueiro.

kfranqueiro:
FYI, I will be adding feedback but also resolving it on a personal branch for merging.

src/Map.ts
```diff
@@ -0,0 +1,160 @@
+import { hasClass } from './decorators';
+import { is } from './object';
+import global from './global';
```
kfranqueiro, line 3: No longer used

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export default class Map<K, V> {
+	private _keys: K[] = [];
```
kfranqueiro, line 6: Change to `protected` so that if someone wants to make an extension, it can access these

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+	/**
+	 * Returns an array of two-value tuples in the form
+	 * of [key, value] in order of insertion.
+	 *
+	 * @return an array of entries in order of insertion
+	 */
+	entries(): any[] {
+		var entries: any[] = [];
+		this._keys.forEach(function(key: K, index: number) {
+			entries.push([key, this.get(key)]);
```
kfranqueiro, line 86: You added the `index` parameter but didn't actually use it here like I suggested.  `this.get(key)` becomes `this._values[index]` to skip an extra invocation.

I'm considering switching this to a straight-up for loop, while I'm at it, actually.

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+	 * element key, and the associated Map instance.
+	 *
+	 * @param callback The function to execute for each map entry,
+	 * @param context The value to use for `this` for each execution
+	 * of the calbackv
+	 */
+	forEach(callback: (key: K, value: V, mapInstance: Map<K, V>) => any, context?: {}) {
+		// don't use this.entries to avoid second forEach call
+		this._keys.forEach(function(key, index) {
+			callback.call(context, key, this.get(key), this);
```
kfranqueiro, line 103: My comments for `entries` are largely applicable here too, actually.

kfranqueiro, line 103: Also, the callback is supposed to be passed the value, then the key, then the Map, but is currently being passed key first.

tests/unit/Map.ts
```diff
@@ -0,0 +1,255 @@
...
+		}
+	},
+
+	iteration: {
+		before() {
+			function foo() {}
+			let object = Object.create(null);
+			let array = new Array();
+			mapArgs = [
+				[0, 0],
```
kfranqueiro, line 229: Missing spaces inside brackets.

tests/unit/Map.ts
```diff
@@ -0,0 +1,255 @@
...
+				[2, object],
+				[3, array],
+				[4, foo],
+				[5, undefined]
+			]
+			map = new Map<number, any>(mapArgs);
+		},
+
+		'callback arguments'() {
+			map.forEach(function (key, value, mapInstance) {
```
kfranqueiro, line 240: key and value are reversed from the spec.

tests/unit/Map.ts
```diff
@@ -0,0 +1,255 @@
...
+			map = new Map<void, void>();
+			assert.deepEqual(map.keys(), []);
+		},
+
+		'nonempty map'(){
+			function foo() { }
+			let object = Object.create(null);
+			let array = new Array();
+			map = new Map<any, any>([
+				[0, 0],
```
kfranqueiro, line 174: Missing spaces inside brackets.

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+		return this._indexOfKey(this._keys, key) > -1;
+	}
+
+	/**
+	 * Returns an array of map keys in order of insertion.
+	 *
+	 * @return an array of keys in order of insertion
+	 */
+	keys(): K[] {
+		return this._keys.slice(0);
```
kfranqueiro, line 134: `slice(0)` is equivalent to just `slice()`.

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+	 *
+	 * @constructor
+	 *
+	 * @param iterable
+	 * An array of two-item tuples used to initially
+	 * populate the map. The first item in each tuple
+	 * corresponds to the key of the map entry. The second
+	 * item corresponds to the value of the map entry.
+	 */
+	constructor(iterable?: any) {
```
kfranqueiro, line 33: Since my comment from a couple of days ago wasn't followed up on, I just looked into it myself.  It is indeed possible to use `array.ArrayLike` here, though we have to forfeit the for...of loop to do so because TS explicitly only allows arrays or strings in it for transpiling to ES5.  Small price to pay for better type information, IMO.

src/Map.ts
```diff
@@ -0,0 +1,160 @@
...
+		if (!(this instanceof Map)) {
```
kfranqueiro, line 34: If the [TS Playground](http://www.typescriptlang.org/Playground#src=class%20MyClass%20%7B%0A%09constructor%28%29%20%7B%0A%09%7D%0A%7D%0A%0Avar%20c%20%3D%20MyClass%28%29%3B) is any indication, TS won't allow you to directly call a class constructor, so this code would be unreachable from TS-compiled code anyway.

tests/unit/Map.ts
```diff
@@ -0,0 +1,255 @@
...
+				[array, 3],
+				[foo, 4]
+			]);
+			assert.deepEqual(map.keys(), [
+				0, '1', object, array, foo
+			]);
+		},
+	},
+
+	values() {
```
kfranqueiro, line 186: We have empty + non-empty for entries/keys now but not values.

tests/unit/all.ts
```diff
@@ -19,3 +19,6 @@ import './string';
 import './UrlSearchParams';
 import './WeakMap';
 import './DateObject';
+import './async/iteration';
```
kfranqueiro, line 22: Not sure how this happened, but these async modules are already listed in all.ts (at the top, in alphabetical order).  Map should also be added in case-insensitive alpha order here.

kfranqueiro:
New mostly-squashed branch at https://github.com/kfranqueiro/core/commits/map (with my cleanup currently separate).  I'll give @bryanforbes a chance to review this if he wants before it gets merged.

kfranqueiro:
TODO: remove entries/keys/values in favor of just using `forEach`.

PR #10, opened 2015-05-14
Implement shim for registering microtasks.

mwistrand:
When an environment does not provide a native means of registering callbacks to the microtask queue, the `queueMicroTask` method will push registered callbacks to an internal queue, which will execute all "microtasks" within a single macrotask, before any other macrotasks have been registered.


kfranqueiro:
Merged by Kris via https://github.com/dojo/core/commit/5df50b58c1df0e06834da49a23c68df971fa06f9.

PR #11, opened 2015-05-14
TransformStream: update unit tests

msssk:
- extract common logic into `beforeEach`
- Ed fixed TransformStream to call transformer methods as methods and preserve context, so the unit tests are no longer sensitive to when transformer methods are mocked or modified.


nicknisi:
Closed via 1184458. Thanks!

PR #13, opened 2015-05-18
Add UrlSearchParams and unit tests

jason0x43:


src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+		}
+		else if (typeof input === 'string') {
+			// Parse the incoming string as a query string
+			list = parseQueryString(input);
+		}
+		else {
+			list = {};
+		}
+
+		Object.defineProperty(this, 'list', { value: list });
```
kfranqueiro: Is this because list is private?  I'm not sure there's much point to it being non-enumerable.  @BryanForbes, want to weigh in?

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * The internal list maps property keys to arrays of values. The value for any property that has been set will be an
+	 * array containing at least one item. Properties that have been deleted will have a value of 'undefined'.
+	 */
+	private list: { [key: string]: string[] };
```
kfranqueiro: I'd suggest underscore-prefixing this, and using `protected`.

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
+import { duplicate } from './lang';
+
+/**
+ * A object with string keys and string or string array values that describes a query string.
```
kfranqueiro: A -> An

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+type ParamList = { [key: string]: string | string[] };
+
+/**
+ * Parse a query string, returning a ParamList object.
```
kfranqueiro: Parse -> Parses (use third-person present for function summaries, as if it is describing what the function does)

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+function parseQueryString(input: string): ParamList {
+	const query = <{ [key: string]: string[] }> {};
+	input.split('&').forEach(entry => {
```
kfranqueiro: Consider `for...of` here instead.  (If we were using `forEach`, we'd avoid fat arrow here since there's no context preservation going on, as per the style guide.)

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+		}
+		else {
+			query[key] = [ value ];
+		}
+	});
+	return query;
+}
+
+/**
+ * An class that represents a URL query search parameters.
```
kfranqueiro: An -> A (or consider dropping "A class that" entirely)

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export default class UrlSearchParams {
+	/**
+	 * Construct a new UrlSearchParams from a query string, an object of parameters and values, or anoter
```
kfranqueiro: Construct -> Constructs, anoter -> another

src/UrlSearchParams.ts
```diff
@@ -0,0 +1,151 @@
...
+		const query = <string[]> [];
+
+		for (let key in this.list) {
+			if (!this.has(key)) {
+				continue;
+			}
+
+			const values = this.list[key];
+			const encodedKey = encodeURIComponent(key);
+			for (let i = 0; i < values.length; i++) {
```
kfranqueiro: Looks like you could use a `for...of` here.

tests/unit/UrlSearchParams.ts
```diff
@@ -0,0 +1,130 @@
...
+				// Assignments after the first will be ignored
+				params = new UrlSearchParams('foo=bar=baz');
+				assert.deepEqual(params.getAll('foo'), [ 'bar' ]);
+
+				// Handle empty keys
+				params = new UrlSearchParams('=foo&');
+				assert.deepEqual(params.getAll(''), [ 'foo', '' ]);
+			},
+
+			UrlSearchParams() {
```
kfranqueiro, line 42: Should we also have a test somewhere to verify that the two instances do not share values?

src/UrlSearchParams.ts
kitsonk: Or using core/WeakMap...

src/UrlSearchParams.ts
jason0x43: `for...of` seems like a good fit. I feel like our style guide is giving up one of the main benefits of arrow functions, though, in not recommending them for their compactness (which is kind of the whole point of the syntax).

src/UrlSearchParams.ts
jason0x43: Apparently grammar had just totally abandoned me by the time I got to block comments.

src/UrlSearchParams.ts
jason0x43: It is.

src/UrlSearchParams.ts
jason0x43: Hmmm...I thought we were trying to stay away from underscore prefixes as a privacy notation, but the style guide recommends it, so I suppose not.

jason0x43: WeakMap is great for privacy, but if we want to allow access in subclasses (which I would assume is why we'd switch from private to protected), it wouldn't work so well.

src/UrlSearchParams.ts
jason0x43: I keep forgetting that exists...

tests/unit/UrlSearchParams.ts
jason0x43, line 42: I added a new 'no data sharing' test for this.

kfranqueiro:
Oops, I neglected to note this was merged in https://github.com/dojo/core/commit/6ec110dc25acc91952422d04d6a2e6c24b5c7c05.  Thanks, and sorry about that!

PR #15, opened 2015-05-19
Date Manipulation and Duration

devpaul:
A Date Object based on https://github.com/dojo/dojo2/blob/master/src/DateObject.ts and a Duration used for describing a period of time.

### DIFF
--- a/src/DateObject.ts
+++ b/src/DateObject.ts
@@ -1,7 +1,9 @@
+import Duration from './Duration';
+
 /**
  * Describes some interval of time used in date calculations
  */
-export interface Interval {
+export interface DateLike {
 	dayOfMonth?: number;
 	hours?: number;
 	milliseconds?: number;
@@ -14,7 +16,7 @@ export interface Interval {
 /**
  * Minimal values needed to describe a date with optional values taken from Interval to define a more specific date
  */
-export interface DateValues extends Interval {
+export interface DateValues extends DateLike {
 	month: number;
 	year: number;
 }
@@ -37,11 +39,6 @@ export interface BasicDate extends DateValues {
 
 var operationOrder = [ 'year', 'month', 'dayOfMonth', 'hours', 'minutes', 'seconds', 'milliseconds' ];
 var days = [ null, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 ];
-var MILLISECONDS_IN_SECOND = 1000;
-var MILLISECONDS_IN_MINUTE = 60000;
-var MILLISECONDS_IN_HOUR = 3600000;
-var MILLISECONDS_IN_DAY = 86400000;
-var MONTHS_IN_YEAR = 12;
 
 function isLeapYear(year: number): boolean {
 	var date = new Date(year, 1, 29);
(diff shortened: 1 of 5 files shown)


src/Duration.ts
```diff
@@ -0,0 +1,54 @@
...
+	constructor(... rest: number[]) {
+		var time = 0;
+		for (var i = 0; i < rest.length; i++) {
+			time += rest[i] * unitValues[i];
+		}
+		this.time = time;
+	}
+
+	get milliseconds(): number {
+		return this.time % Duration.SECOND;
```
bryanforbes, line 24: Why is this doing a mod?

src/Duration.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	get hours(): number {
+		return convertUnits(this.time % Duration.DAY, Duration.HOUR);
+	}
+
+	get days(): number {
+		return convertUnits(this.time, Duration.DAY);
+	}
+
+	add(value: Duration): Duration {
```
devpaul, line 43: include support for add(value: number): Duration

src/Duration.ts
```diff
@@ -0,0 +1,54 @@
...
+		return new Duration(this.time + value.time);
+	}
+
+	difference(value: Duration): Duration {
```
devpaul, line 47: include support for difference(value: number): Duration

src/Duration.ts
```diff
@@ -0,0 +1,54 @@
...
+		return new Duration(value.time - this.time);
+	}
+
+	compare(value: Duration): number {
```
devpaul, line 51: include support for compare(value: number): number

src/Duration.ts
bryanforbes, line 24: I talked with @devpaul and the mod is to make the `duration.milliseconds` API similar to `dateObject.milliseconds` which makes perfect sense.

src/Duration.ts
```diff
@@ -0,0 +1,54 @@
+var unitValues = [1, 1000, 60000, 3600000, 86400000];
+
+function convertUnits(time: number, unitMilliseconds: number) {
+	return (time > 0 ? 1 : -1) * Math.floor(Math.abs(time) / unitMilliseconds);
+}
+
+export default class Duration {
+	public static SECOND = unitValues[1];
```
devpaul, line 8: move public statics into export const enum Interval { Millisecond, Second = 1000, ... }

src/DateObject.ts
```diff
@@ -0,0 +1,351 @@
...
+
+	get dayOfWeek(): number {
+		return this._date.getDay();
+	}
+	get timezoneOffset(): number {
+		return this._date.getTimezoneOffset();
+	}
+
+	add(value: number): DateObject;
+	add(value: Duration): DateObject;
```
devpaul, line 260: We should add DateObject.subtract()

src/DateObject.ts
```diff
@@ -0,0 +1,351 @@
...
+		}
+		else {
+			// Attempt to put value into a DateObject and translate to a time
+			result.time += new DateObject((<any> value)).time;
+		}
+
+		return result;
+	}
+
+	difference(value: number): Duration;
```
devpaul, line 290: Consider renaming difference() to until() to match with Duration.until() and disambiguate from subtract(). difference() is ambiguous in general practice when returning a Duration once the difference goes beyond the border of a month or leap year. until() is more direct in answering date1.until(date2) -> duration.

PR #16, opened 2015-05-20
Encodings: Ascii, Utf8, Hex and Base64

jdonaghue:


src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+		toJSON: function (): string {
+			return JSON.stringify(Array.prototype.slice.call(this));
+		},
+		length: data.length
+	}];
+}
+
+function validateDecodingArgs(data: any) {
+	if (data == null) {
+		return false;
```
kitsonk: Why not `===`?

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	}
+
+	if (data.length == null) {
+		throw new TypeError('Argument must have a length property');
```
kitsonk: Won't this throw weird... should it be `if ('length' in data) {`?

src/encoding.ts
jdonaghue: I opted to use == here so that it will catch undefined as well

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+		let decoded = '';
+
+		for (let i = 0, length = data.length; i < length; i++) {
+			let byte1 = data[i] & 0xFF;
+
+			if ((byte1 & 0x80) === 0) {
+				decoded += decodeUtf8EncodedCodePoint(byte1);
+			}
+			else if ((byte1 & 0xE0) === 0xC0) {
+				let byte2 = data[++i] & 0xFF;
```
kitsonk, line 168: Why are these being hoisted?  Wouldn't it be better to declare them above instead of conditional hoisting?

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	 *
+	 * @param data
+	 * The byte buffer to be decoded
+	 */
+	static decode(data: ByteBuffer): string {
+		if (!validateDecodingArgs(data)) {
+			return '';
+		}
+
+		let decoded = '';
```
kitsonk, line 159: I looked at the style guide, and it doesn't address this sort of implicit typing, but I don't like it.  I would much rather we were explicitly typing wherever reasonable, because it makes intent so much easier.

tests/unit/encoding.ts
```diff
@@ -0,0 +1,251 @@
...
+	{
+		encoding: '\uDFFF',
+		decoding: [0xED, 0xBF, 0xBF]
+	}
+];
+
+registerSuite({
+	name: 'Encoding Classes',
+
+	'Ascii': {
```
kitsonk: No explicit tests for `validateDecodingArgs()`.  We should try to have negative tests as well as positive tests, especially when there is code that is specifically validating arguments, we should test for invalid arguments.

src/encoding.ts
kfranqueiro: If there's no expectation that data would be explicitly null or undefined (only not defined at all), then that would make more sense here.

src/encoding.ts
kfranqueiro, line 168: What do you mean by "hoisted"?  `let` has block scope.

src/encoding.ts
kfranqueiro, line 159: Good point about it not being in the style guide, but it was a conscious agreement during earlier discussions.

I don't really see the point to `let decoded: string = '';`, since you can obviously tell that `''` is already a string.

The one possible instance where I'd consider enforcing explicit typing even where unnecessary is for return types of exported functions.

src/encoding.ts
kitsonk, line 168: I didn't realise ES6 let was block scope.  Learned something new.  Though obviously the emit to ES5, this wouldn't be block scope though, right?  Only when emitted ES6.  I guess TS would do any guarding against mis-scoping?

src/encoding.ts
kitsonk, line 159: Fair enough it was agreed... maybe adding to the style guide that explicit agreement.

I would still argue that `let decode: string = '''` still makes the intent explicitly clear, and you can only get away with the with primitives (IIRC).

src/encoding.ts
kfranqueiro, line 168: Yeah, that's the idea.  TS also ensures that variable names are unique if they need to be (e.g. if you have one block-scoped `i` variable shadowing another one in an outer block scope).

src/encoding.ts
kfranqueiro, line 159: I have a local change to the style guide to add this, though I'm hoping we merge the contributing and style guides into one repo rather than having to update it in 4 or more of them.

Actually type inferencing happens on objects too...and those are the times it _can_ bite you.  Primitives are where it seems particularly excessive.

src/encoding.ts
jdonaghue: Good suggestion, not sure why I wrote it the way I did

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	return true;
+}
+
+function validateUtf8EncodedCodePoint(codePoint: number): void {
+	if ((codePoint & 0xC0) !== 0x80) {
+		throw Error('Invalid continuation byte');
+	}
+}
+
+function decodeUtf8EncodedCodePoint(codePoint: number, validationRange: number[] = [0, Infinity], checkSurrogate?: boolean): string {
```
kfranqueiro: Space after `[` and before `]` in default value

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	if (codePoint < validationRange[0] || codePoint > validationRange[1]) {
+		throw Error('Invalid continuation byte');
+	}
+
+	if (checkSurrogate && codePoint >= 0xD800 && codePoint <= 0xDFFF) {
```
kfranqueiro, line 17: We were going to make these constants either here or in `string`, right?

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	toJSON?: () => string;
+	length: number;
+}
+
+/**
+ * Ascii codec that provides facilities for encoding a string (or string coerced object)
+ * into an Ascii encoded byte buffer as well as decoding an Ascii encoded byte buffer
+ * into a string.
+ */
+export class Ascii {
```
kfranqueiro: - Is there a point to these being classes rather than objects with 2 functions?
- Should there be an interface that these implement/conform to?

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	 * @param data
+	 * Any object that is either a string or can be coerced into a string
+	 */
+	static encode(data: any): ByteBuffer {
+		let buffer: ByteBuffer;
+
+		[ data, buffer ] = normalizeEncodingArgs(data, Ascii);
+
+		for (let i = 0, length = data.length; i < length; i++) {
+			buffer[i] = data.charCodeAt(i);
```
kfranqueiro, line 59: Is this technically accurate, given that `charCodeAt` returns a 16-bit numeric Unicode value?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+	 * @param data
+	 * Any object that is either a string or can be coerced into a string
+	 */
+	static encode(data: any): ByteBuffer {
+		let buffer: ByteBuffer;
+
+		[ data, buffer ] = normalizeEncodingArgs(data, Hex);
+
+		for (let i = 0, length = data.length; i < length; i++) {
+			let encodedChar = parseInt(data.charCodeAt(i).toString(16), 16);
```
kfranqueiro: Can you explain to an encoding noob (me) what the `parseInt` and `toString(16)` here accomplishes?  It appears to me that they'd just cancel each other out.

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+
+		return decoded;
+	}
+}
+
+/**
+ * Base64 codec that provides facilities for encoding a string (or string coerced object)
+ * into an Base64 encoded byte buffer as well as decoding an Base64 encoded byte buffer
+ * into a string.
+ * Adapted from http://www.nczonline.net/blog/2009/12/08/computer-science-in-javascript-base64-encoding/
```
kfranqueiro: This should actually be able to use `atob` and `btoa` in all but IE9 (possibly with [some extra considerations for Unicode](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/btoa#Unicode_Strings)).

Meanwhile, Dojo 1 already has base64 code that we can lift from (though the code is unfortunately harder to read...)

kfranqueiro:
Have we considered having tests that do encode -> decode and/or decode -> encode to ensure that the value is the same as the original after running through both operations?  (I'm not sure offhand if any of the existing tests are effectively doing that.)

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+		encoded += String.fromCharCode(codePoint >>> 0x10 & 0x03FF | 0xD800);
+		codePoint = 0xDC00 | codePoint & 0x03FF;
+	}
+
+	encoded += String.fromCharCode(codePoint);
+
+	return encoded;
+}
+
+export interface ByteBuffer {
```
kfranqueiro: @jason0x43 may have more detailed thoughts on this, but it seems like it'd be far more useful (though probably more difficult) to be able to return something that is useful in the respective platform this is used.

At the _very_ least, it seems like it'd be more useful for this to actually be an array, not just an array-like.

src/encoding.ts
kfranqueiro, line 59: dojox/encoding in Dojo 1 already has an `ascii85` module which looks quite different...  However, I'm not sure whether that's actually what we're looking for here.

https://en.wikipedia.org/wiki/Ascii85

src/encoding.ts
```diff
@@ -0,0 +1,379 @@
...
+			}
+		}
+
+		return decoded;
+	}
+}
+
+/**
+ * Hex codec that provides facilities for encoding a string (or string coerced object)
+ * into an Hex encoded byte buffer as well as decoding an Hex encoded byte buffer
```
jason0x43, line 213: The original goal was to convert a hex encoded string, like "DEADBEEF01", into a byte buffer, and vice versa, rather than to store a string in a particular binary format.

src/encoding.ts
jdonaghue: `Is there a point to these being classes rather than objects with 2 functions?`
There isn't really a great reason for these being classes over plain old objects. I opted for classes though because they can be extended. So, if someone were to want to subclass any of these they would be able to. I know thats not an amazing argument but when comparing the two options a class based approach is the more flexible of the two. That said I am not married to it and could easily be convinced that an object based approach is better

`Should there be an interface that these implement/conform to?`
There isn't an interface because all of the methods are static.

src/encoding.ts
jdonaghue, line 17: yes, thank you for reminding me!

src/encoding.ts
kitsonk: Well if you are going to test for any falsey value, why not `!data` then?

src/encoding.ts
```diff
@@ -0,0 +1,347 @@
+const BASE64_KEYSTR = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=';
+
+function normalizeEncodingArgs(data: any): [ string, number[] ] {
```
jason0x43: Since we're not using a custom ByteBuffer class, this method doesn't seem to be accomplishing much. We should just have the encoding functions only accept strings.

src/encoding.ts
```diff
@@ -0,0 +1,347 @@
...
+ * into a string.
+ */
+export class Ascii {
+	/**
+	 * Encode a string or string coerced object into an Ascii byte buffer
+	 *
+	 * @param data
+	 * Any object that is either a string or can be coerced into a string
+	 */
+	static encode(data: any): number[] {
```
jason0x43: Is there any reason not to make the signature `encode(data: string)`?

src/encoding.ts
```diff
@@ -0,0 +1,347 @@
...
+		return buffer;
+	}
+
+	/**
+	 * Decode an Ascii encoded byte buffer into a string
+	 *
+	 * @param data
+	 * The byte buffer to be decoded
+	 */
+	static decode(data: number[]): string {
```
jason0x43: This would be more convenient as `decode(data: Uint8Array | Buffer | number[])` since Uint8Array and Buffer act the same as far as `decode` is concerned (so the same code path would work for all of them). You could use a type alias, like `type ByteBuffer = Uint8Array | Buffer | number[]`, giving `decode(data: ByteBuffer)`.

src/encoding.ts
jason0x43, line 159: As long as the compiler can unambiguously figure things out, is there any benefit to including unnecessary explicit types?

src/encoding.ts
```diff
@@ -0,0 +1,347 @@
...
+	}
+}
+
+/**
+ * Utf8 codec that provides facilities for encoding a string (or string coerced object)
+ * into an Utf8 encoded byte buffer as well as decoding an Utf8 encoded byte buffer
+ * into a string.
+ * Inspired by the work of: https://github.com/mathiasbynens/utf8.js
+ */
+export class Utf8 {
```
jason0x43: The codecs should implement a `Codec` interface.

src/encoding.ts
kfranqueiro, line 159: See number 2 under https://github.com/dojo/guidelines/blob/master/STYLE.md#typescript for how I worded it.  Basically I'm only advocating explicit types where necessary or in exported functions or class methods.

src/encoding.ts
jdonaghue: You can't have static methods on an interface in Typescript unfortunately.

src/encoding.ts
jdonaghue: ok good idea I will make this change

src/encoding.ts
jason0x43: I think the benefits of using 2-function modules with an interface outweigh the ability to subclass. Having an interface will allow these to be easily passed into functions, which will be a key use of codecs. If someone wants to extend one of these codecs, they can just wrap it and call the wrapped object's methods.

jdonaghue: ok that sounds reasonable. I will make that change as well.

src/encoding.ts
jason0x43, line 159: Hmmm...I don't really think adding the type improves clarity much; you're going to know what it should be the first time you try to compile the code, and your IDE is always going to be able to tell you. It could probably help enforce the public API, though. Explicitly setting a return type will prevent you from accidentally returning something invalid (which will probably only come up in library code, but hey, that's what we're writing).

src/encoding.ts
```diff
@@ -0,0 +1,337 @@
...
+	encode(data: string): number[];
+	decode(data: ByteBuffer): string;
+}
+
+/**
+ * Ascii codec that provides facilities for encoding a string
+ * into an Ascii encoded byte buffer as well as decoding an Ascii encoded byte buffer
+ * into a string.
+ */
+export const Ascii: Codec = {
```
jason0x43: The codec names should use lower camel case now since they're not classes. (And they're not constants in the sense that they should be all-capped).

kfranqueiro: Just to avoid any potential confusion here... you're not saying `const` is unwarranted here, right?  You're just saying that they shouldn't be all-caps either?

tests/unit/encoding.ts
```diff
@@ -0,0 +1,235 @@
...
+		'.encode()'() {
+			let buffer = Ascii.encode(ENCODED_STRING);
+			assert.strictEqual(buffer.length, ENCODED_STRING.length);
+
+			for (let i = 0, length = buffer.length; i < length; i++) {
+				assert.strictEqual(buffer[i], ASCII_BUFFER[i]);
+			}
```
jason0x43: Since we're outputting number arrays all the time now, the test could just use assert.deepEqual instead of checking the length and individual character values. Also, there's an assert.lengthOf for checking the length of things.

src/encoding.ts
kfranqueiro, line 159: Yeah, ensuring you don't accidentally return something unexpected is definitely part of why I put it in there.  Also just that it's generally easier to eyeball the first line of the function to get all the type information out of it if you're reading it.

tests/unit/encoding.ts
```diff
@@ -0,0 +1,200 @@
...
+			assert.deepEqual(buffer, ASCII_BUFFER);
+
+			buffer = ascii.encode('1');
+			assert.lengthOf(buffer, 1);
+			assert.strictEqual(buffer[0], 49);
+		},
+
+		'.decode()'() {
+			let decoded = ascii.decode(ASCII_BUFFER);
+			assert.lengthOf(decoded, ENCODED_STRING.length);
```
jason0x43: There's no need for a lengthOf check when the next thing is check for strict equality.

tests/unit/encoding.ts
```diff
@@ -0,0 +1,200 @@
...
+	}
+];
+
+registerSuite({
+	name: 'Encoding Classes',
+
+	'ascii': {
+		'.encode()'() {
+			let buffer = ascii.encode(ENCODED_STRING);
+			assert.lengthOf(buffer, ENCODED_STRING.length);
```
jason0x43: There's no need for a lengthOf check when the next thing is checking for deep equality (which includes length).

src/encoding.ts
```diff
@@ -0,0 +1,341 @@
...
+	/**
+	 * Encode a string into an Ascii byte buffer
+	 *
+	 * @param data
+	 * A text string to be encoded
+	 */
+	encode(data: string): number[] {
+		let buffer: number[] = [];
+
+		for (let i = 0, length = data.length; i < length; i++) {
```
jason0x43, line 58: There should be a check that data isn't null/undefined. If it is null, the encoder should just return an empty array.

src/encoding.ts
```diff
@@ -0,0 +1,341 @@
...
+			 * Surrogates
+			 * http://en.wikipedia.org/wiki/Universal_Character_Set_characters
+			 */
+			if (encodedChar > HIGH_SURROGATE_MIN && encodedChar < HIGH_SURROGATE_MAX) {
+				let lowSurrogate = data.charCodeAt(++i);
+				if (lowSurrogate >= LOW_SURROGATE_MIN && lowSurrogate < LOW_SURROGATE_MAX) {
+					encodedChar = 0x010000 + (encodedChar - 0xD800) * 0x0400 + (lowSurrogate - 0xDC00);
+				}
+				else {
+					i--;
```
jason0x43: Directly modifying the loop-control variable inside a for loop seems...odd. Maybe that's just me.

jdonaghue: would it be preferable for me to change this to a while-loop?

kfranqueiro: I think this would be acceptable as a `for` loop if you increment `i` only _inside_ the `if` (at which point you can remove the `else` entirely).  One control var tweak is better than two.  (You'd have to do `i + 1` on the line before the `if` though.)

A while loop wouldn't really help, since you'd have to explicitly increment twice in some cases.

src/encoding.ts
```diff
@@ -0,0 +1,337 @@
+type ByteBuffer = Uint8Array | Buffer | number[];
+const BASE64_KEYSTR = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/=';
+const HIGH_SURROGATE_MIN = 0xD800;
```
kfranqueiro: My original suggestion was to put these surrogate constants in `string.ts` then import them here (and also use them there).

src/encoding.ts
jdonaghue: I have made this change

src/encoding.ts
```diff
@@ -0,0 +1,333 @@
...
+ * into a string.
+ */
+export const ascii: Codec = {
+	/**
+	 * Encode a string into an Ascii byte buffer
+	 *
+	 * @param data
+	 * A text string to be encoded
+	 */
+	encode(data: string = ''): number[] {
```
jason0x43: This doesn't handle the case when data is null, only when it's undefined. Use something like `validateDecodingArgs` inside the function instead.

src/encoding.ts
```diff
@@ -0,0 +1,333 @@
...
+		let buffer: number[] = [];
+
+		for (let i = 0, length = data.length; i < length; i++) {
+			let encodedChar = data.charCodeAt(i);
+
+			/**
+			 * Surrogates
+			 * http://en.wikipedia.org/wiki/Universal_Character_Set_characters
+			 */
+			if (encodedChar > HIGH_SURROGATE_MIN && encodedChar < HIGH_SURROGATE_MAX) {
```
jason0x43: Should this range be inclusive (>= and <= instead of > and <)?

src/encoding.ts
```diff
@@ -0,0 +1,333 @@
...
+				let lowSurrogate = data.charCodeAt(i + 1);
+				if (lowSurrogate >= LOW_SURROGATE_MIN && lowSurrogate < LOW_SURROGATE_MAX) {
```
jason0x43: Should the `LOW_SURROGATE_MAX` comparison be inclusive?

src/encoding.ts
```diff
@@ -0,0 +1,333 @@
...
+ */
+export const utf8: Codec = {
+	/**
+	 * Encode a string into an Utf8 byte buffer
+	 *
+	 * @param data
+	 * A text string to be encoded
+	 */
+	encode(data: string = ''): number[] {
+		let buffer: number[] = [];
```
kitsonk: Ok, might be another thing with me getting my head wrapped around ES6, but why isn't this a `const`?  You are never going to reassign this or modify the pointer in any way.

src/encoding.ts
jason0x43: Actually, why not use `string.codePointAt` for this?

src/encoding.ts
```diff
@@ -0,0 +1,333 @@
...
+	 */
+	decode(data: ByteBuffer): string {
+		if (!validateDecodingArgs(data)) {
+			return '';
+		}
+
+		let decoded = '';
+
+		for (let i = 0, length = data.length; i < length; i++) {
+			decoded += parseInt('' + data[i], 10).toString(16).toUpperCase();
```
jason0x43: Why use `parseInt` instead of just `data[i].toString(16)`? `data[i]` should already be a number.

src/encoding.ts
kfranqueiro: RE using `codePointAt`, that was an idea I had brought up, but I wasn't sure if it'd add complexity regarding knowing how many characters to skip.

And yeah, I'd suspect this should be inclusive, not exclusive.

tests/unit/encoding.ts
```diff
@@ -0,0 +1,204 @@
...
+	'ascii': {
+		'.encode()'() {
+			let buffer = ascii.encode(ENCODED_STRING);
+			assert.deepEqual(buffer, ASCII_BUFFER);
+
+			buffer = ascii.encode('1');
+			assert.lengthOf(buffer, 1);
+			assert.strictEqual(buffer[0], 49);
+
+			assert.deepEqual([], ascii.encode(undefined));
```
jason0x43, line 77: Also test `null`

tests/unit/encoding.ts
```diff
@@ -0,0 +1,204 @@
...
+		},
+
+		'.decode()'() {
+			let decoded = ascii.decode(ASCII_BUFFER);
```
jason0x43, line 82: It'd be good to add tests using Buffers and Uint8Arrays (on compatible platforms) since a decoder can theoretically accept those as well as numeric arrays.

src/encoding.ts
kfranqueiro: I think that's a fair point here and was just overlooked.  Easy to miss those things with objects that get mutated but not reassigned.

src/encoding.ts
jdonaghue: The only problem with using `codePointAt` is that I need to keep track of what position I am at in the source string and that would be difficult with that utility function because it doesn't return any indicator of whether a low surrogate was provided or not. 

Also, according to this: http://people.mozilla.org/~jorendorff/es6-draft.html#sec-string.prototype.codepointat both of these ranges should both be exclusive which if thats correct then both this and `codePointAt` are wrong currently

src/encoding.ts
jdonaghue: yeah good point, i will change

src/encoding.ts
jdonaghue: I was using `parseInt` because it was originally returning a collection of `numbers`. Now it is returning a string so `parseInt` doesn't make sense. Good catch!

tests/unit/encoding.ts
jdonaghue, line 82: Good idea I will add some `Buffer` and `Uint8Array` tests, thanks!

src/encoding.ts
jdonaghue: duh! I read that wrong, you are right they should be inclusive. I will change

kfranqueiro: IINM, I think you're reading the spec backwards.  Step 9 tests for things being _outside_ of the range, in which case the endpoints are exclusive.  We're testing for them being _inside_ the range here and in `string.ts`.  (The difference between `n > max || n < min`, and `n <= max && n >= min`)

tests/unit/encoding.ts
```diff
@@ -0,0 +1,246 @@
...
+
+			assert.strictEqual(ascii.decode(undefined), '');
+			assert.strictEqual(ascii.decode(null), '');
+
+			if (has('host-node')) {
+				let buffer = new Buffer('The cat and the hat');
+				assert.strictEqual(ascii.decode(buffer), 'The cat and the hat');
+			}
+			else if (has('host-browser')) {
+				let buffer = new Uint16Array(ASCII_BUFFER);
```
jason0x43, line 95: This fails in IE9. We should add a has test for typed arrays and use that here.

src/encoding.ts
```diff
@@ -0,0 +1,349 @@
...
+export const ascii: Codec = {
+	/**
+	 * Encode a string into an Ascii byte buffer
+	 *
+	 * @param data
+	 * A text string to be encoded
+	 */
+	encode(data: string): number[] {
+		if (!validateArguments(data)) {
+			data = '';
```
jason0x43: You could just return `[]` here.

tests/unit/encoding.ts
```diff
@@ -0,0 +1,246 @@
...
+			assert.strictEqual(ENCODED_STRING, ascii.decode(ascii.encode(ENCODED_STRING)));
+
+			assert.strictEqual(ascii.decode(undefined), '');
+			assert.strictEqual(ascii.decode(null), '');
+
+			if (has('buffer')) {
+				let buffer = new Buffer('The cat and the hat');
+				assert.strictEqual(ascii.decode(buffer), 'The cat and the hat');
+			}
+			else if (has('uint16Array')) {
```
jason0x43, line 94: Any particular reason to use uint16 here instead of uint8? The values in the test buffers all fit in 8-bits, and the Codec interface is defined as accepting a Uint8Array.

src/encoding.ts
```diff
@@ -0,0 +1,349 @@
...
+ */
+export const ascii: Codec = {
+	/**
+	 * Encode a string into an Ascii byte buffer
+	 *
+	 * @param data
+	 * A text string to be encoded
+	 */
+	encode(data: string): number[] {
+		if (!validateArguments(data)) {
```
jason0x43: Since `validateArguments` is just testing for data to be null or undefined, just say

``` js
if (data == null) {
   ..
}
```

and get rid of `validateArguments`.

tests/unit/encoding.ts
jdonaghue, line 94: Because the test string contains characters higher than an octect supports. 65399 and  65382 wont fit in 8-bits

src/encoding.ts
jdonaghue: good call, I will do this thanks!

tests/unit/encoding.ts
kfranqueiro, line 94: Isn't ASCII 8-bit?  Why would we be testing with 16-bit numbers?

Also, I'm not sure `has` tests are warranted for things that we're only using in tests, not implementation.

kfranqueiro:
I have an [encoding-cleanup](https://github.com/kfranqueiro/core/commits/encoding-cleanup) branch which I hope to merge later today or tomorrow.  The main things addressed were:
- exporting the `ByteBuffer` type since other APIs will want to reference it (e.g. crypto)
- jsdoc comment tweaking
- removing the `has` tests this added (since they're only used in unit tests, not implementations)
- using separate strings for ascii and utf8 tests
- using Uint8Array where it makes sense rather than Uint16Array in tests
- replacing more values corresponding to the new surrogate constants

kfranqueiro:
Merged in https://github.com/dojo/core/commit/8b70efadfaea3e67df52743dceb7b18af64fae80.

PR #20, opened 2015-05-22
Streams - add EventedStreamSource implementation

msssk:
The `EventedStreamSource` module enables creation of a `ReadableStream.Source` from an evented object (Dojo `Evented` or DOM node).

### DIFF
--- a/src/streams/ReadableStream.ts
+++ b/src/streams/ReadableStream.ts
@@ -37,7 +37,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's start operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	start?: (controller: ReadableStreamController<T>) => Promise<void>;
+	start?(controller: ReadableStreamController<T>): Promise<void>;
 
 	/**
 	 * Requests that source enqueue chunks.  Use the controller to close the stream when no more chunks can
@@ -46,7 +46,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's pull operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	pull?: (controller: ReadableStreamController<T>) => Promise<void>;
+	pull?(controller: ReadableStreamController<T>): Promise<void>;
 
 	/**
 	 * Indicates the stream is prematurely closing and allows the source to do any necessary clean up.
@@ -54,7 +54,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's pull operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	cancel?: (reason?: any) => Promise<void>;
+	cancel?(reason?: any): Promise<void>;
 }
 
 /**
--- a/src/streams/ReadableStream.ts
+++ b/src/streams/ReadableStream.ts
@@ -37,7 +37,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's start operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	start(controller: ReadableStreamController<T>): Promise<void>;
+	start?: (controller: ReadableStreamController<T>) => Promise<void>;
 
 	/**
 	 * Requests that source enqueue chunks.  Use the controller to close the stream when no more chunks can
@@ -46,7 +46,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's pull operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	pull(controller: ReadableStreamController<T>): Promise<void>;
+	pull?: (controller: ReadableStreamController<T>) => Promise<void>;
 
 	/**
 	 * Indicates the stream is prematurely closing and allows the source to do any necessary clean up.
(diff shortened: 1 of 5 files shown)


src/streams/adapters/EventedStreamSource.ts
```diff
@@ -0,0 +1,50 @@
...
+	protected _controller: ReadableStreamController<Event>;
+	protected _target: EventTargetTypes;
+	protected _events: Array<ExtensionEvent | string>;
+	protected _handles: Handle[];
+
+	constructor(target: EventTargetTypes, type: EventTypes) {
+		this._target = target;
+
+		if (Array.isArray(type)) {
+			this._events = <any> type;
```
kitsonk, line 21: Why the any cast?

src/streams/adapters/EventedStreamSource.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Source } from '../ReadableStream';
+import ReadableStreamController from '../ReadableStreamController';
+
+type EventTargetTypes = Evented | EventEmitter | EventTarget;
+type EventTypes = Array<ExtensionEvent | string> | ExtensionEvent | string;
+
+export default class EventedStreamSource implements Source<Event> {
+	protected _controller: ReadableStreamController<Event>;
+	protected _target: EventTargetTypes;
+	protected _events: Array<ExtensionEvent | string>;
```
kitsonk, line 14: Why not make a type alias of `type EventNames = ExtensionEvent | string` and then use `EventNames[]`?

src/streams/adapters/EventedStreamSource.ts
```diff
@@ -0,0 +1,50 @@
...
+	protected _handles: Handle[];
+
+	constructor(target: EventTargetTypes, type: EventTypes) {
+		this._target = target;
+
+		if (Array.isArray(type)) {
+			this._events = <any> type;
+		}
+		else {
+			this._events = [ <any> type ];
```
kitsonk, line 24: Again, why the any cast?

tests/unit/streams/adapters/EventedStreamSource.ts
```diff
@@ -0,0 +1,76 @@
...
+		let appleEvent = {
+			type: 'apple',
+			test: 'value'
+		};
+		let orangeEvent = {
+			type: 'orange',
+			test: 'value'
+		};
+
+		source = new EventedStreamSource(emitter, [ 'apple', 'orange' ]);
```
kitsonk, line 49: Is there any purpose in declaring this outside of the block, assigning it in beforeEach and then reassigning it here?

tests/unit/streams/adapters/EventedStreamSource.ts
```diff
@@ -0,0 +1,76 @@
...
+		let enqueueCallCount = 0;
+
+		stream.controller.enqueue = function (chunk: Event) {
+			enqueueCallCount += 1;
+		};
+
+		source.cancel();
+		emitter.emit(testEvent);
+		assert.strictEqual(enqueueCallCount, 0, 'Canceled source should not call controller.enqueue');
+	}
```
kitsonk, line 75: No tests for ExtensionEvent being passed to the constructor.  Also missing any tests for non-Evented emitters.

src/streams/ReadableStream.ts
```diff
@@ -37,7 +37,7 @@ export interface Source<T> {
 	 * @returns A promise that resolves when the source's start operation has finished.  If the promise rejects,
 	 * 		the stream will be errored.
 	 */
-	start(controller: ReadableStreamController<T>): Promise<void>;
+	start?: (controller: ReadableStreamController<T>) => Promise<void>;
```
edhager: This is the same signature conversion that is not necessary and inconsistent with the rest of the code.  This one should be:
`start?(controller: ReadableStreamController<T>): Promise<void>;`

tests/unit/streams/adapters/EventedStreamSource.ts
msssk, line 75: I believe these will have to be functional tests.

src/streams/adapters/EventedStreamSource.ts
msssk, line 21: Necessary for compilation; TS support for union types is still somewhat crude.

src/streams/adapters/EventedStreamSource.ts
msssk, line 24: Necessary for compilation; TS support for union types is still somewhat crude.

src/streams/adapters/EventedStreamSource.ts
msssk, line 14: Stylistic preference I guess.

`EventTargetTypes` is used in multiple places, so worth creating a type alias for.
`EventTypes` is a long garbled mess, so worth creating a type alias for (to keep the code on line 17 somewhat nicer looking)

Line 14 is the only place this union type is used, and it's short enough that it looks good inline.

tests/unit/streams/adapters/EventedStreamSource.ts
msssk, line 49: It was done like this in anticipation of having more tests that used the same value for source... but in the current suite, it is indeed unnecessary - there are only two test cases that use source, and they each use a different value.

msssk, line 49: Actually, source is used (indirectly) in multiple tests, so I'm leaving this code the way it is.

src/streams/ReadableStream.ts
msssk: fixed in https://github.com/msssk/core/commit/e01f17873d101eb384999251f17e6f0dcb478982

src/streams/adapters/EventedStreamSource.ts
kitsonk, line 24: I think what we have agreed is that we should at least comment why we have an `any` cast, because the intent isn't always (seldom) clear.

PR #22, opened 2015-05-26
Streams: miscellaneous fixes, tests & jsdoc updates

msssk:


src/streams/WritableStream.ts
```diff
@@ -22,70 +18,82 @@ export interface Record<T> {
...
 	/**
 	 * Indicates the stream is prematurely closing due to an error.  The sink should do any necessary cleanup
-	 * and release resources.
+	 * and release resources. When a stream calls `abort` it will discard any queued chunks. If the sink does not
+	 * provide an `abort` method then the stream will call `close` instead.
+	 *
 	 * @param reason The reason the stream is closing.
 	 */
-	abort(reason?: any): Promise<void>;
+	abort?: (reason?: any) => Promise<void>;
```
edhager: Why did you change the style of the signatures for all of the methods in Sink?  It is not consistent with how we are defining interfaces in other places.  I think this one should be:
`abort?(reason?: any): Promise<void>;`

src/streams/WritableStream.ts
```diff
@@ -257,18 +263,23 @@ export default class WritableStream<T> {
...
+			);
 		}
 
 		if (this.state === State.Closed) {
 			// 4.2.4.4-2
 			return Promise.resolve();
 		}
-
-		if (this.state === State.Errored) {
+		else if (this.state === State.Errored) {
```
edhager: Since the previous `if` returns, an else is not needed here.  We have this style of allowing each precondition check to be its own `if` statement.

src/streams/WritableStream.ts
```diff
@@ -277,23 +288,32 @@ export default class WritableStream<T> {
...
+				new Error('WritableStream method called in context of object that is not a WritableStream instance')
+			);
 		}
 
-		if (this.state === State.Errored) {
+		// 4.2.4.5-2
+		if (this.state === State.Closed) {
+			return Promise.reject(new TypeError('Stream is already closed'));
+		}
+		else if (this.state === State.Closing) {
```
edhager: Same here, the `else` is not needed.

src/streams/WritableStream.ts
msssk: I don't think the syntax you proposed works. When I changed the methods to being optional, I tried that and it didn't work, so I ended up with the somewhat odd syntax you see now.

msssk: I checked the history, and indeed I changed to that syntax when I made the optional. I tried the syntax you suggested and it does work now, so I've pushed an update with that change.

src/streams/WritableStream.ts
msssk: Thanks, fixed!

src/streams/WritableStream.ts
```diff
@@ -104,7 +112,7 @@ export default class WritableStream<T> {
 	protected _queue: SizeQueue<Record<T>>;
 	protected _writing: boolean;
 
-	constructor(underlyingSink: Sink<T>, strategy: Strategy<T> = {}) {
+	constructor(underlyingSink: Sink<T> = {}, strategy: Strategy<T> = {}) {
```
edhager, line 115: Why did you add a default value for underlyingSink?  The code is written to handle an undefined underlyingSink.  It uses util.invokeOrNoop to take care of that.  Was this change made for some scenario that isWriteableStream needs to return true when a WritableStream is instantiated with no sink?  On line 210, the sink is set to undefined which is not consistent with it having a default value.

src/streams/WritableStream.ts
edhager: `else` not needed

src/streams/WritableStream.ts
edhager: `else` not needed

src/streams/WritableStream.ts
```diff
@@ -277,23 +288,32 @@ export default class WritableStream<T> {
...
 
-		if (this.state === State.Errored) {
+		// 4.2.4.5-2
+		if (this.state === State.Closed) {
+			return Promise.reject(new TypeError('Stream is already closed'));
+		}
+		else if (this.state === State.Closing) {
+			return Promise.reject(new TypeError('Stream is already closing'));
+		}
+		else if (this.state === State.Errored) {
```
edhager: `else` not needed

src/streams/WritableStream.ts
```diff
@@ -304,25 +324,35 @@ export default class WritableStream<T> {
...
+				new Error('WritableStream method called in context of object that is not a WritableStream instance')
+			);
 		}
 
-		if (this.state === State.Errored) {
+		// 4.2.4.6-2
+		if (this.state === State.Closed) {
+			return Promise.reject(new TypeError('Stream is closed'));
+		}
+		else if (this.state === State.Closing) {
```
edhager: `else` not needed

src/streams/WritableStream.ts
```diff
@@ -304,25 +324,35 @@ export default class WritableStream<T> {
...
+			return Promise.reject(new TypeError('Stream is closing'));
+		}
+		else if (this.state === State.Errored) {
```
edhager: `else` not needed

src/streams/WritableStream.ts
msssk, line 115: The spec indicates it should default to an empty object:
https://streams.spec.whatwg.org/#ws-class-definition

I know there's not much practical use to it, but without a default value, you can still create a WritableStream, but if you don't provide a sink the constructor throws an unintelligible error (I ran into this while writing tests).

If we are not going to provide a default, then we should at least check the params to see if a sink was provided to the constructor and abort gracefully if not.

msssk, line 115: Line 210 sets the sink to undefined to release the WritableStream's reference to the sink. At that point it no longer matters what sink is set to - calling any method will fail due to checks for `state === State.Closed`

edhager:
Running the tests via `grunt test` results in a request test failing.  Back to msssk to fix it.

msssk:
@edhager Fixed the `request/node` unit test failure in https://github.com/msssk/core/commit/bdb7813414b4bc53e4201f74da4f6a247b5103f1

Issue #3, opened 2015-05-10
Typings Bundle does not include the whole library...


kitsonk:
When I use:

``` sh
grunt dist
```

I get `dist/typings/dojo-core/dojo-core-2.0.0.d.ts` output and it includes all the modules in the root of `src`, but not the modules in the sub directories `src`.  For example, the file has:

``` ts
import ObjectObserver from 'dojo-core/observers/ObjectObserver';
```

but does not have a corresponding `declare module 'dojo-core/observers/ObjectObserver'`.

While this comes back as a valid file within the repository, when copying `dist` to another repository, it generates several TypeScript errors about not being able to resolve the modules.

I don't know if this is a `dts-generator` issue, or something to do with the configuration of it within `core`.

bryanforbes:
I committed a change (fd2df9157cc1eb73c9859ac28a19dc65b3772c17) that fixes the exported interfaces so the typings are correctly generated. There's also a [pull request for dts-generator](https://github.com/SitePen/dts-generator/pull/18) which will throw an error when the situation you're seeing arises within the code. Should we keep this open until a new version of dts-generator is released with the fix?

kitsonk:
Well, I can confirm it resolves the specific issue I reported. Technically
the change to dts-generator is an improvement. So I say close.

kitsonk:
This appears to be resolved.

PR #23, opened 2015-05-27
Fix broken type exports

jason0x43:
- Fix some unexported type issues that weren't being indicated by dts-generator. They will be indicitated by the next version of dts-generator.
- <s>Update Scheduler to allow QueueItems to be scheduled directly -- this will allow subclasses to use custom QueueItems to annotate scheduled callbacks.</s>

References dojo/dom#12

### DIFF
--- a/src/Scheduler.ts
+++ b/src/Scheduler.ts
@@ -35,7 +35,12 @@ export default class Scheduler {
 	 */
 	queueFunction: (callback: (...args: any[]) => any) => Handle;
 
-	protected _defer(item: QueueItem): Handle {
+	protected _defer(callback: (...args: any[]) => void): Handle {
+		const item: QueueItem = {
+			isActive: true,
+			callback: callback
+		};
+
 		if (!this._deferred) {
 			this._deferred = [];
 		}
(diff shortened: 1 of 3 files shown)


src/Scheduler.ts
```diff
@@ -94,15 +89,22 @@ export default class Scheduler {
...
+		let item: QueueItem;
+
+		if (typeof callback === 'function') {
+			item = {
+				isActive: true,
+				callback: <(...args: any[]) => void> callback
+			};
+		}
+		else {
+			item = <QueueItem> callback;
```
bryanforbes: There is currently no test for this condition. Please write one.

jason0x43:
Hmmm...I'm not terribly happy about exposing QueueItem via `schedule`. There should be an internal method that does what `schedule` does (including deferring if necessary) but that only accepts QueueItems. The external API should still only accept callbacks. Reworking (and adding tests)...

kfranqueiro:
I've merged this to master in https://github.com/dojo/core/commit/75704bd17220c17d818abf6c6d0c4270754eaaba.

I have to wonder though...  is this technically a deficiency in dts-generator?  I mean, the things that are exported in this changeset are really internal to the modules in question, and if you were consuming the ts files in this repository directly, I don't think there would be a problem.

IIUC, the problem is that if they're not exported, dts-generator can't pick them up, but still defines other interfaces as inheriting from the unexported ones?  Should it be capable of negotiating that itself, or is the limitation due to language services or something?

jason0x43:
The problem is that if an exported interface uses unexported interfaces, dts-generator won't include any of them (explicitly exported or not) in its generated file, but will appear to complete successfully.

kfranqueiro:
Oh, so it's not that it just exposes "holey" interfaces, it doesn't end up exposing them at all...  But is that due to a limitation of its own, or due to not being able to get information from TS language services about them at all?

kitsonk:
@kfranqueiro the behaviour of dts-generator is issue #3 but also I don't know if we updated to the dts-generator that was merged on the back of SitePen/dts-generator#21 or not.

jason0x43:
dts-generator adds typings to a typings file using `Program.emit` from TypeScript's language services. If some are missing, there's nothing dis-generator can directly do about it.

The change in SitePen/dts-generator#21 does something similar to tsc's `noEmitOnError` flag. When that flag is set, the compiler will check that no diagnostic messages were generated for a particular emit operation, and will block the emit (and fail) if any were. Without the flag, the compiler will still emit code even when diagnostic messages are generated.

PR #24, opened 2015-05-27
Remove EventTarget interface since TS already has one

kfranqueiro:
I noticed that we have our own `EventTarget` interface defined in `on.ts`, but TS already has one of its own:

``` ts
interface EventTarget {
    addEventListener(type: string, listener: EventListenerOrEventListenerObject, useCapture?: boolean): void;
    dispatchEvent(evt: Event): boolean;
    removeEventListener(type: string, listener: EventListenerOrEventListenerObject, useCapture?: boolean): void;
}
```

The main differences are that TS's interface has `dispatchEvent` while ours doesn't, and ours has `accessKey` (though I'm not entirely sure why) while TS's doesn't (it exists on `HTMLElement` in TS's typings).

I figured I'd throw this PR out there in case it's worth the cleanup.  If there is a legitimate reason why we were doing what we're doing, feel free to just close this.  But as far as I can tell, nothing was explicitly relying on our definition and tests still pass.

### DIFF
--- a/src/on.ts
+++ b/src/on.ts
@@ -11,12 +11,6 @@ export interface EventEmitter {
 	removeListener(event: string, listener: EventCallback): EventEmitter;
 }
 
-export interface EventTarget {
-	accessKey?: string;
-	addEventListener(event: string, listener: EventCallback, capture?: boolean): void;
-	removeEventListener(event: string, listener: EventCallback, capture?: boolean): void;
-}
-
 interface DOMEventObject extends EventObject {
 	bubbles: boolean;
 	cancelable: boolean;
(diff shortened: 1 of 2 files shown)


bryanforbes:
I'm not sure why `accessKey` is on that interface, but I'd rather not require `dispatchEvent` for registering events.

kfranqueiro:
As per discussion elsewhere, the no-dispatchEvent argument seems a tad weak when you consider that `on` also accepts `EventEmitter` or `Evented` objects.  However, I'm willing to leave this be until we further explore the necessity of extension events.

bryanforbes:
:+1: Go ahead and merge.

PR #25, opened 2015-05-27
Remove unneeded checks from Promise, update tests

jason0x43:


tests/unit/Promise.ts
```diff
@@ -76,19 +70,33 @@ export function addPromiseTests(suite: any, Promise: PromiseType) {
...
+
+			race() {
+				let dfd = this.async();
+				let iterable: any[] = [];
+
+				iterable[0] = Promise.resolve(0);
+				iterable[3] = Promise.resolve(3);
+
+				Promise.race(iterable).then(dfd.callback(function (value: number) {
+					assert.strictEqual(value, 0);
```
kfranqueiro: I may just be uninformed and paranoid here, but is this a sufficient test?
- The array contains two already-resolved promises, separated by two unpopulated indices
- The first item (before any undefined indices) is the answer anyway

I suppose `race` is still going to iterate over the entire array.  What would be the behavior of this if we didn't handle sparse arrays though?  Would this assertion fail or would the entire operation throw or hit the errback (which we're not handling in this test)?

jason0x43: Yeah, it would make more sense to set index 1 instead of 0.

I would assume a `race` that didn't handle sparse arrays correctly would simply not skip unset indices. It would access indices 1 and 2 in the current test and treat the resulting `undefineds` as meaningful values. In the current test that shouldn't matter since the first index is set. If we set index 1 instead of 0, though, then a bad `race` would resolve with `undefined`.

jason0x43: Interesting. `Promise.race` and `Promise.all` operate on iterables, and ES6 array iterators don't do anything special for sparse arrays. In Chrome and FF, the following iterates over the undefined values as well as the explicitly set values:

``` js
var q = [];
q[2] = 2;
q[4] = 4;
var i = q[Symbol.iterator]();
var n = i.next();
while (!n.done) {
    console.log(n.value);
    n = i.next();
}
```

Promise should be updated to match.

jason0x43:
Since native promises don't treat sparse arrays specially, I removed this behavior from Promise. I also updated the tests to better test sparse array handling as suggested by Ken, which exposed the issue.

PR #29, opened 2015-05-29
Add webcrypto has test

jason0x43:
References dojo/crypto#4, dojo/crypto#7

### DIFF
--- a/src/has.ts
+++ b/src/has.ts
@@ -77,3 +77,4 @@ add('weakmap', function () {
 	}
 	return false;
 });
+add('webcrypto', typeof global.SubtleCrypto !== 'undefined');


kfranqueiro:
I could be totally ignorant on this, but what's the rationale for this going into core rather than a `has` module in `crypto`?

kitsonk:
This is a debate we had a lot in Dojo 1.  Where do we put the hasAdd for features...  I still vote that it should go into the most logical module that needs it first.  Even when we have a has module loader, we should try to not "pollute" the main dojo/has with too many features and in a lot of ways the core shim module shouldn't be dynamically loaded anyways, though it may dynamically load modules.

kfranqueiro:
We decided that within each package, all feature tests would go into a central `has` module, to make the feature tests a lot easier to find to avoid the where's-waldo problem that Dojo 1 had.  The question I'm asking here is specifically around which package's `has` module it goes into.

kitsonk:
Ok, that makes sense...  It should go into dojo-crypto/has.  Also, why test for SubtleCrypto?  SubtleCrypto is only part of "Webcrypto".  [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto) indicates this:

``` js
var cryptoObj = window.crypto || window.msCrypto; // for IE 11
```

So I would argue more precise feature detection, though I will admit I haven't looked in detail of how dojo/crypto is offloading to native so it maybe appropriate.

jason0x43:
Ah. I remembered long ago having a has test in async (when that was a thing) that Bryan moved to core, so I figured we were grouping that way. I'll move it to crypto.

As for why test for SubtleCrypto, the plan was to only focus on WebCrypto at the moment (hence the  `has('webcrypto')`). We could potentially add other in-browser providers, but those could also be other features.

Issue #26, opened 2015-05-28
lang needs something that Just Works for simple value copying


kfranqueiro:
I know we are already mulling over how to split up `lang.copy` to simplify its API and also the mental overhead regarding the combinations it supports.  On top of that, one thing it doesn't seem to support is simply copying values, a la `lang.mixin` in Dojo 1.

Here's an interesting (though probably rather edgy) case I just ran into (trying to use it during development of `dom/geometry`, https://github.com/dojo/dom/issues/8) that affects various browsers differently, presumably due to how they treat native `ClientRect` objects:
1. Get a native `ClientRect` object via `someElement.getBoundingClientRect()`
2. Attempt to coerce to a plain ol' object by running through `copy`

Results:
- On Chrome this happens to Just Work as expected - you get an object with the expected properties with numeric values.
- On Firefox (tested 38) and IE (tested 9 and 11), you end up with seemingly an empty object.

Tangentially, I would definitely vouch for not burying `sources` in a `kwArgs` object if we can help it, and accommodating one source without needing an array.  Having to do that in this case seems rather obtuse.  (I realize that other derivative APIs like `create` and `duplicate` wrap this complexity away.)

kfranqueiro:
WIP on https://github.com/kfranqueiro/dom/commits/geometry, currently with basic support + tests for the get APIs.

Dojo 1's `dom-geometry.position` had an optional flag which would adjust based on the document's scroll position.  It did not, however, adjust based on scroll position of any child elements.  Need to think through what is necessary/best for this API.

kitsonk:
I just took a look at it, maybe should have before... That isn't really a usable API in my opinion.

To do the most basic with it, you have to do this:

``` ts
let copy = lang.copy({ sources: [ orig ] });
```

Wow!

I mean of course I could have done this, in most cases:

``` ts
let copy = Object.create(orig);
```

Some of the functionality in the `kwArgs` might be better split out as seperete API functions:

``` ts
export function copy(...sources: any[]): any;
export function copyDeep(...sources: any[]): any;
```

I also don't understand the use case where we wouldn't want to copy descriptors.  I could see there maybe variations in how we determine what keys to copy over, but why would we not copy the descriptor?

kitsonk:
Looks like @kfranqueiro specific use case in this is interesting... The challenge is the a ClientRec instances has no own properties.  All the properties are in the prototype and they are all "odd".

[...]

PR #31, opened 2015-06-09
Simplify lang's object copying (addresses issue #26: https://github.com/dojo/core/issues/26)

mwistrand:
- Remove inheritance checking from lang.copy.
- Break lang.copy into lang.assign, lang.assignDeep, lang.copy, and lang.copyDeep.

### DIFF
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -10,95 +10,55 @@ function isObject(item: any): boolean {
 	return item && typeof item === 'object' && !Array.isArray(item) && !(item instanceof RegExp);
 }
 
-function copyArray(array: any[], kwArgs: CopyArgs): any[] {
+function copyArray(array: any[], inherited: boolean): any[] {
 	return array.map(function (item: any): any {
 		if (Array.isArray(item)) {
-			return copyArray(item, kwArgs);
+			return copyArray(item, inherited);
 		}
 
-		return isObject(item) ?
-			copy({
+		return !isObject(item) ?
+			item :
+			_mixin({
+				deep: true,
+				inherited: inherited,
 				sources: [ item ],
-				deep: kwArgs.deep,
-				descriptors: kwArgs.descriptors,
-				inherited: kwArgs.inherited,
-				assignPrototype: kwArgs.assignPrototype
-			}) :
-			item;
+				target: {}
+			});
 	});
 }
 
-export function copy(kwArgs: CopyArgs): any {
-	const sources: Hash<any>[] = kwArgs.sources;
-	let target: any;
+interface MixinArgs {
+	deep: boolean;
+	inherited: boolean;
+	sources: {}[];
+	target: {};
+}
 
-	if (!sources.length) {
-		throw new RangeError('lang.copy requires at least one source object.');
-	}
+function _mixin(kwArgs: MixinArgs): {} {
+	const deep = kwArgs.deep;
+	const inherited = kwArgs.inherited;
+	const target = kwArgs.target;
 
-	if (kwArgs.assignPrototype) {
-		// create from the same prototype
-		target = Object.create(Object.getPrototypeOf(sources[0]));
-	}
-	else {
-		// use the target or create a new object
-		target = kwArgs.target || {};
-	}
+	for (let source of kwArgs.sources) {
+		for (let key in source) {
+			if (inherited || hasOwnProperty.call(source, key)) {
+				let value: any = (<any> source)[key];
 
-	for (let source of sources) {
-		if (kwArgs.descriptors) {
-			// if we are copying descriptors, use to get{Own}PropertyNames so we get every property
-			// (including non enumerables).
-			const names = (kwArgs.inherited ? getPropertyNames : Object.getOwnPropertyNames)(source);
-
-			for (let name of names) {
-				// get the descriptor
-				const descriptor = (kwArgs.inherited ?
-					getPropertyDescriptor : Object.getOwnPropertyDescriptor)(source, name);
-				let value = descriptor.value;
-
-				if (kwArgs.deep) {
+				if (deep) {
 					if (Array.isArray(value)) {
-						descriptor.value = copyArray(value, kwArgs);
+						value = copyArray(value, inherited);
 					}
 					else if (isObject(value)) {
-						descriptor.value = copy({
-							sources: [ value ],
+						value = _mixin({
 							deep: true,
-							descriptors: true,
-							inherited: kwArgs.inherited,
-							assignPrototype: kwArgs.assignPrototype
+							inherited: inherited,
+							sources: [ value ],
+							target: {}
 						});
 					}
 				}
 
-				// and copy to the target
-				Object.defineProperty(target, name, descriptor);
-			}
-		}
-		else {
-			// If we aren't using descriptors, we use a standard for-in to simplify skipping
-			// non-enumerables and inheritance. We could use Object.keys when we aren't inheriting.
-			for (let name in source) {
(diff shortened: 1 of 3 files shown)


kfranqueiro:
While this looks simpler, it still seems to be more complex than what @bryanforbes had suggested earlier in the week, which was:
- assign = for...in + hasOwnProperty
- copy = for...in

i.e. we'd only concern ourselves with enumerable properties, and I don't think any of the property descriptor stuff would be necessary at all anymore.  (Personally, I'm wondering if we can come up with a clearer name for `copy` to make the distinction more evident though...)

Also, what's the rationale for e.g. `assignDeep` rather than `deepAssign`?  The latter sounds more natural.  The main thing I see with the former is it keeps both functions together alphabetically.

kfranqueiro:
As per discussion:
- These APIs won't concern themselves with property descriptors.  If the need for that arises, we will consider additional APIs.
- `copy` will become `mixin`, matching Dojo 1's `lang.mixin`
- `assignDeep` will become `deepAssign`, `copyDeep` will become `deepMixin`

mwistrand:
I amended my commit with these changes, and also replaced `Array#reduce` with `for` loops to help improve performance.

src/lang.ts
```diff
@@ -10,95 +10,55 @@ function isObject(item: any): boolean {
...
+	const target = kwArgs.target;
 
-			for (let name of names) {
-				// get the descriptor
-				const descriptor = (kwArgs.inherited ?
-					getPropertyDescriptor : Object.getOwnPropertyDescriptor)(source, name);
-				let value = descriptor.value;
+	for (let source of kwArgs.sources) {
+		for (let key in source) {
+			if (inherited || hasOwnProperty.call(source, key)) {
```
kfranqueiro, line 44: Why isn't this just `source.hasOwnProperty(key)` (and remove the `hasOwnProperty` variable)?  (Seems like a remnant of older code, not sure if the older code had any more reason to be doing this so awkwardly...)

src/lang.ts
```diff
@@ -10,95 +10,55 @@ function isObject(item: any): boolean {
...
+function copyArray(array: any[], inherited: boolean): any[] {
 	return array.map(function (item: any): any {
 		if (Array.isArray(item)) {
-			return copyArray(item, kwArgs);
+			return copyArray(item, inherited);
 		}
 
-		return isObject(item) ?
-			copy({
+		return !isObject(item) ?
```
kfranqueiro, line 19: Is there a reason this was reversed?  Typically I expect non-negated conditions to read easier.

src/lang.ts
```diff
@@ -10,95 +10,55 @@ function isObject(item: any): boolean {
...
+				deep: true,
+				inherited: inherited,
 				sources: [ item ],
-				deep: kwArgs.deep,
-				descriptors: kwArgs.descriptors,
-				inherited: kwArgs.inherited,
-				assignPrototype: kwArgs.assignPrototype
-			}) :
-			item;
+				target: Object.create(null)
```
kfranqueiro: I feel like I'm forgetting something here, but what benefit does this have over just passing an empty object?

src/lang.ts
```diff
@@ -120,25 +80,37 @@ export function create(prototype: {}, ...mixins: {}[]): {} {
...
 	});
 }
 
+export function duplicate(source: {}): {} {
+	const target = Object.create(Object.getPrototypeOf(source));
+
+	return deepMixin(target, source);
+}
+
+// TODO: This is no longer used. Should it be removed?
```
kfranqueiro: I vote to remove.

src/lang.ts
```diff
@@ -160,6 +132,8 @@ export function getPropertyNames(object: {}): string[] {
 	return names;
 }
 
+// TODO ObjectObserver has a similar method that goes a step further by providing a default descriptor.
+// Should this implementation be replaced with that one, as this one is not used anymore?
```
kfranqueiro: #40 discusses moving the observe stuff to its own file anyway, so does this become entirely removable then?  If so, I'd vote to remove.

src/lang.ts
```diff
@@ -10,95 +10,55 @@ function isObject(item: any): boolean {
...
-		target = Object.create(Object.getPrototypeOf(sources[0]));
-	}
-	else {
-		// use the target or create a new object
-		target = kwArgs.target || {};
-	}
+interface MixinArgs {
+	deep: boolean;
+	inherited: boolean;
+	sources: {}[];
```
kfranqueiro, line 33: This signature seems like an improvement over `CopyArgs` (which had `any`), but I'm wondering if there's anything better we can do.  AFAICT, this will match Arrays, Dates, and seemingly even primitives (though I honestly don't understand why).

I tried using `Hash<any>` in its place, but that makes passing valid object hashes more difficult in cases where implicit typing takes effect, e.g.:

``` ts
assign(target, { foo: 42 }); // compiles

let obj = { foo: 42 };
assign(target, obj); // doesn't compile - needs explicit type to work
```

Any thoughts?  (cc @bryanforbes)

tests/unit/lang.ts
```diff
@@ -8,104 +8,107 @@ import { Es5Observer, Es7Observer } from 'src/observers/ObjectObserver';
...
+		const source = Object.create({ a: 1 }, {
+			b: {
+				enumerable: false,
+				configurable: true,
+				writable: true,
+				value: 2
+			}
 		});
+		source.c = 3;
+
```
kfranqueiro, line 22: You appear to be testing for shallow copying in the `mixin` test but not in the `assign` test.

kfranqueiro:
+1 to better organization in the benchmark module.

It seems like we're missing jsdoc comments for a lot of these functions.  Could you add them?

src/lang.ts
neonstalwart, line 44: doing it this way is to guard against someone redefining `hasOwnProperty` on the object.  it ensures that this code uses the `hasOwnProperty` it expects to be using.

src/lang.ts
mwistrand, line 19: Personal preference; I don't want to have to look over multiple object keys before being able to figure out which variables are actually involved in the ternary.

src/lang.ts
mwistrand: I've been doing this out of habit, but I'm guessing having a blank slate isn't always going to be what other developers will want. I'll change it to `{}`.

src/lang.ts
mwistrand, line 33: Interestingly, native `Object.assign` doesn't discriminate. For example, run `Object.assign(new Date(), /\s/)` in Firefox. I haven't tested whether it does anything useful, but at the very least it doesn't throw an error. As for the typing, I personally think it's fine. It's explicit that objects are meant to be used, and it doesn't require the user to throw in additional coercions just to use it.

mwistrand:
@kfranqueiro I'm in the process of adding the jsdoc comments now. Given that @bryanforbes is removing `lang.observe`, I will leave those comments out.

mwistrand:
`lang.ts` is now updated with the following changes:
- Remove `lang.getPropertyNames` and `lang.getPropertyDescriptor`.
- Rename `assignDeep` to `deepAssign`, `copy` to `mixin`, and `copyDeep` to `deepMixin`.
- Add jsdoc comments to all `lang` methods, save `observe` as it will be removed from the package.

kfranqueiro:
I have this ready to merge on a local branch, but I want to look at what merging #40 in addition looks like before I push buttons.

kfranqueiro:
It occurs to me that at this point, `object.assign` can go away.  I'm going to look at its tests to see if I should port any into the `lang` test suite.

PR #32, opened 2015-06-10
Fix: string.escapeRegExp should escape "/".

mwistrand:


edhager:
Merged in 224a8779039de84fd85ebc9ca23191adeb38a1dc

PR #33, opened 2015-06-12
Remove extension event support

### LINKED ISSUE #30: Extension events
We've had a couple of discussions back and forth on the topic of extension events.  I'm opening this issue to document some of the considerations that have come up so far.
- [Extension events in Dojo 1](http://dojotoolkit.org/reference-guide/1.10/dojo/on.html#extension-events) can be kind of hard for devs to wrap their heads around (and hard for us to teach), since it involves providing a function as an event "type", and that function is in turn responsible for listening for any relevant events and firing the listener that is passed when appropriate
- The built-in extension events that exist in Dojo 1 aren't likely to be needed in Dojo 2:
  - The Pointer Events Polyfill is expected to be used in place of `dojo/touch`
  - All modern browsers support `mouseenter` and `mouseleave`
  - `a11yclick` ideally shouldn't need to exist (see https://github.com/dojo/dom/issues/10)
- That said, there are still cases where essentially "composite" events can be useful; for example, to respond to an enter keypress or a click on a row in a grid:
  
  ``` js
  function clickOrEnter(node, listener) {
      return on(node, 'click,keyup', function (event) {
          if (event.keyCode === 13 || !('keyCode' in event)) {
              listener.call(this, event);
          }
      });
  }
  
  on(document.getElementById("test"), clickOrEnter, ...);
  ```
- Another idea that has been tossed around by one or two people is instituting a registry for custom string event type names.
- One concern that has been brought up in the past is whether something could be done or a pattern promoted to minimize the number of event listeners that are actually hooked up for extension events - i.e., have a document-level handle for each type, to avoid numerous handlers being registered each time an extension event is used
- Possibly a separate consideration, but we also need to support utilities like `once` and `pausable` somehow.  While it would probably be possible to implement them via Dojo 1 style extension events, one could argue that that's muddying the waters of extension events' intended use (and personally, I'd rather see them implemented in a way that they can be used with `aspect` too).

I'm starting to throw around some examples in https://gist.github.com/kfranqueiro/b61a0d4d1939063842b2 just to illustrate different approaches.

kfranqueiro:
We currently can't think of any strong use cases for this, so we're
removing it at least for the time being.

Resolves #30.

### DIFF
--- a/src/on.ts
+++ b/src/on.ts
@@ -17,10 +17,6 @@ export interface EventTarget {
 	removeEventListener(event: string, listener: EventCallback, capture?: boolean): void;
 }
 
-export interface ExtensionEvent {
-	(target: any, listener: EventCallback, capture?: boolean): Handle;
-}
-
 interface DOMEventObject extends EventObject {
 	bubbles: boolean;
 	cancelable: boolean;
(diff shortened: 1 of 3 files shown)


bryanforbes:
:+1: Go ahead and merge.

PR #39, opened 2015-06-15
Load implementation

### LINKED ISSUE #35: Add API to dynamically load modules
Because ES6 module imports can only be used at the top-level of a module (they cannot be used inside of a block) and because Node's loader does not understand AMD loader plugins, it would be good to have an API for dynamically loading modules. The API should take a variable number of module IDs and return a promise that fulfills with an array of the module values. At this time, I think this should go into its own module and be called `require`. The implementation could be as simple as the following:

``` ts
import Promise from './Promise';

declare var require: Function;

const _require: {
    (...moduleIds: string[]): Promise<any[]>;
} = (function () {
    if (typeof define === 'function' && define.amd) {
        return function (...moduleIds: string): Promise<any[]> {
            return new Promise(function (resolve, reject) {
                // TODO: hook up reject
                require(moduleIds, resolve);
            });
        };
    }
    else if (has('host-node')) {
        return function (...moduleIds: string): Promise<any[]> {
            return Promise.resolve(moduleIds.map(function (moduleId) {
                return require(moduleId);
            }));
        };
    }
    return function () {
        return Promise.reject(new Error('Unknown loader'));
    };
})();
export { _require as default };
```

Once we have this, the code at https://github.com/dojo/core/blob/master/src/request.ts#L69-L82 can be removed for a `dojo-core/require` call. In addition, the observers pulled in in `observe.ts` could be done this way as well as any of the providers in crypto and bundles in i18n.

bryanforbes:
Fixes #35.


src/load.ts
```diff
@@ -0,0 +1,75 @@
+import has from './has';
+import Promise from './Promise';
+
+declare var require: Function;
+declare var define: any;
```
kfranqueiro: I have noticed we've been adding these two declarations in various places (actually, usually when I see us declare `define`, it includes the `amd` key).  Should we centralize them somewhere?  (Am I wrong to be concerned about clobbering/conflicts?)

Edit: Actually, I suppose the one place I saw it in core was in request, which this refactors.  I saw it in the crypto API PR in that repo.

kitsonk: And do we need an interface somewhere, or do we need a typings from the loader that we should depend on?

src/load.ts
```diff
@@ -0,0 +1,75 @@
...
+export interface Load {
+	(module: ModuleObject, ...moduleIds: string[]): Promise<any[]>;
+}
+
+export interface ModuleObject {
+	id: string;
+}
+
+function getBaseId(moduleObject: ModuleObject): string {
+	var id: string = moduleObject.id;
```
kitsonk: `let`?

src/load.ts
```diff
@@ -0,0 +1,75 @@
...
+	var index: number = id.lastIndexOf('/');
```
kitsonk: `const`?

src/load.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const load: Load = (function () {
+	if (!hostLoad) {
+		return function () {
+			return Promise.reject(new Error('Unknown loader'));
+		};
+	}
+	else {
+		return function (module: ModuleObject, ...moduleIds: string[]): Promise<any> {
+			var baseId = getBaseId(module);
```
kfranqueiro: I'm not convinced this is going to work for Node modules...  It seems to me that Node module IDs report absolute paths and include the file extension.  Moreover, they also seem to include backslash rather than forward slash on Windows.

It makes no sense to me that Node's `module.id` reports the filename, but that seems to be what it does, and even [what it's documented to do](https://nodejs.org/api/modules.html#modules_module_id).

src/load.ts
```diff
@@ -0,0 +1,75 @@
...
+})();
+
+const load: Load = (function () {
+	if (!hostLoad) {
+		return function () {
+			return Promise.reject(new Error('Unknown loader'));
+		};
+	}
+	else {
+		return function (module: ModuleObject, ...moduleIds: string[]): Promise<any> {
```
kfranqueiro: Would it make more sense to pass `require` rather than `module` here, which would be a context require in the case of AMD?  (Especially given Node.js's apparent absurdity with what it reports `module.id` as, see next comment)

bryanforbes: I chose `module` instead of `require` because, according to the AMD spec, there's no way to get the absolute module ID given a relative module ID. This means that while passing `require` from Node and both the Dojo 1 and 2 loaders would work fine, `require` from RequireJS would be pretty much useless.

kfranqueiro: Am I missing something?  I was thinking that if the user passes context require, you have no need for the absolute module ID anymore, because the thing you're using it for right now is to resolve relative paths manually.

src/load.ts
```diff
@@ -0,0 +1,75 @@
...
+			moduleIds = moduleIds.map(function (moduleId) {
+				if (moduleId[0] === '.') {
+					moduleId = baseId + '/' + moduleId.substring(2);
+				}
+				return moduleId;
+			});
+			return hostLoad(moduleIds);
+		};
+	}
+})();
```
kitsonk, line 59: I am missing the "value" of the dual wrap of the `hostLoad` and the `load` in separate closures.  I mean the code is right, but it seems to add another layer that appears to be somewhat unnecessary, especially because hostLoad doesn't get exported.

kitsonk:
Test cases? :wink:

src/load.ts
bryanforbes: I think I became dumb yesterday and conflated too many things. Passing `require` is the best option. I'll update the pull request.

src/load.ts
bryanforbes, line 59: I didn't want to repeat the code for calculating the absolute module IDs, but it's not necessary if `require` is passed instead of `module`. Ignore this.

src/load.ts
bryanforbes: I'm not sure if we need to provide typings for loader internals. The problem with providing those is that the typings for node's `require` and AMD's `require` are a bit different. Since we're compiling to UMD, most of the need for loader typings is reduced and this module's default function reduce the need for those typings even further.

bryanforbes:
I'll write some test cases tonight. It'll be interesting to test the Node path since Intern runs in an AMD loader...

bryanforbes:
I finished up the test cases. There's not a ton to check. One thing to note: the coverage data is off in Node because we're using RequireJS which adds some extra stuff to the interpreted file which throws off istanbul. This is fixed once we switch to the new loader.

tests/unit/load.ts
```diff
@@ -0,0 +1,54 @@
+import assert = require('intern/chai!assert');
```
kitsonk, line 1: `import * as assert from 'intern/chai!assert';`

kitsonk:
Would it be useful for the tests to also load text resources, instead of just ~~JSON~~ JavaScript Objects? (and actually test loading some JSON too)

Also, I get the reason why, but it is a shame the API is `load(require, ...mids);`  Is there no create way to figure out how to automatically shorten it to `load(...mids);` or make a context require optional in some way?

tests/unit/load.ts
bryanforbes, line 1: The `assert` module is not an ES6 module, so it (more than likely) shouldn't be imported with ES6 semantics. This may have been fixed in later builds of TS (I've seen instances where `import * as ...` isn't callable), but it's still confusing since in ES6, `assert` would be a non-callable object.

bryanforbes:
I just pushed a change (and rebased on master) so the `require` argument is an optional first argument. I added tests for both contextual and global-only loading for both Node and AMD.

With regards to text/JSON loading, this module is meant to be a bridge to the module loader. In Node, that means you can actually pull in JSON objects with it, but in the browser (AMD) you can't and you'd have to use a plugin.

src/load.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface AMDRequire {
+	(moduleIds: string[], callback: (...modules:any[]) => void): void;
+}
+export interface NodeRequire {
+	(moduleId: string): any;
+}
+export type Require = AMDRequire | NodeRequire;
+
+export interface Load {
+	(require: Require|string, ...moduleIds: string[]): Promise<any[]>;
```
kitsonk: why not?

``` typescript
export interface Load {
    (require: Require, ...moduleIds: string[]): Promise<any[]>;
    (...modulesIds: string[]): Promise<any[]>;
}
```

tests/unit/load.ts
kitsonk, line 1: Ok.

src/load.ts
bryanforbes: Off the top of my head, I don't remember why I did this. Perhaps I was being clever. When I review this over the weekend, I'll see if I can come up with a better defense ;).

kitsonk: I think you implementation logic is 100% correct.  I just think the exposed API could be a bit cleaner (though it is technically correct).  Let's just admit, the availability of union types of blinded us from thinking of overloading things!  I have found myself getting wound into knots with union types only to realises that overloading actually makes things cleaner.

bryanforbes:
Merged in aa982d927b0a1fc481367bdae6c7927553525b6d.

PR #40, opened 2015-06-15
Move lang.observe and observer implementations to their own file

bryanforbes:
This moves `lang.observe` out of lang (as previously discussed) and moves the observer classes into the same file.


src/observe.ts
```diff
@@ -0,0 +1,330 @@
...
+}
+
+export interface ObserveArgs {
+	listener: (events: PropertyEvent[]) => any;
+	nextTurn?: boolean;
+	onlyReportObserved?: boolean;
+	target: {}
+}
+
+export interface Observer extends Handle {
```
kfranqueiro, line 320: Thoughts on moving these interfaces up since they're currently used above?

src/lang.ts
```diff
@@ -171,9 +168,12 @@ export function getPropertyDescriptor(object: Object, property: string): Propert
 }
 
 export function isIdentical(a: any, b: any): boolean {
```
kfranqueiro, line 170: Since you seem to have moved the logic from `object.is` here, should we also consider what to do with that module?  It seems like it will be removable once we also create `assign` and `deepAssign` in `lang`.

Meanwhile, I'm actually not sure whether the `-0` code is even necessary; `-0 === +0` is already `true` in JS.

bryanforbes, line 170: Yes, once the copying stuff lands in master, I don't see any reason for `src/object.ts` to exist. The only reason the Observer classes were using `object.is` was to prevent a circular reference with `lang`. Since it's all been moved to one module in `observe`, I updated `lang.isIdentical` to be more in line with what `Object.is` does.

As we discussed in the call, the changes for `-0` and `+0` are to support `lang.isIdentical(-0, +0) === false`.

src/observe.ts
bryanforbes, line 320: I was trying to preserve alphabetical order, which means `ES5` and `ES7` need to move before `KwArgs`...

src/observe.ts
```diff
@@ -0,0 +1,330 @@
+import has from './has';
+import { Handle, Hash } from './interfaces';
+import { getPropertyDescriptor, isIdentical } from './lang';
```
kfranqueiro, line 3: `ObjectObserver.ts` used to have its own version of `getPropertyDescriptor` which also included default behavior if no descriptor yet existed.  The version in `lang` doesn't have that behavior.  Is this change intentional or should this be preserving the behavior of `getPropertyDescriptor` as it existed in `ObjectObserver`?

(Incidentally, #31 was going to remove `getPropertyDescriptor` from lang anyway, so I'd probably opt to keep the internal function in this module.)

tests/unit/observers/ObjectObserver.ts
```diff
@@ -1,248 +0,0 @@
...
-			testDestroy.call(this, Es7Observer);
-		}
-	},
-
-	'Es5Observer': {
-		'propertyChanges': function () {
-			testPropertyChanges.call(this, Es5Observer);
-		},
-
-		'when attempting to observe an unwritable property': function () {
```
kfranqueiro, line 90: I don't seem to see any equivalent to this test within the new test suite. Is that intentional?

kfranqueiro:
I was hoping to merge this along with #31, but I still have a couple of questions for @bryanforbes on this.  I don't think it should conflict too much - the main thing is that `getPropertyDescriptor` will no longer exist in `lang` after #31 is merged (and either way, that didn't have some of the logic that `ObjectObsever`'s function of the same name had - see also https://github.com/dojo/core/pull/40#discussion_r34257050).

tests/unit/observe.ts
```diff
@@ -0,0 +1,142 @@
...
+			a: {
+				enumerable: false,
+				configurable: true,
+				writable: true,
+				value: 1
+			}
+		});
+
+		let b: number;
+		let length: number;
```
devpaul, line 82: `b` and `length` are unused

src/observe.ts
```diff
@@ -0,0 +1,330 @@
...
+			if (this.nextTurn) {
+				if (!this._currentlyScheduled) {
+					this._currentlyScheduled = {};
+					scheduler.schedule(this._boundDispatch);
+				}
+
+				this._currentlyScheduled[property] = event;
+			}
+			else {
+				this._listener([ event ]);
```
devpaul, line 104: Can we add a test to cover this scenario?

The test should create an observer with nextTick = false; observe a property and be notified of a change synchronously before the test finishes.

src/observe.ts
```diff
@@ -0,0 +1,330 @@
...
+				let descriptor: PropertyDescriptor;
+				if (property in target) {
+					descriptor = getPropertyDescriptor(target, property);
+				}
+				else {
+					descriptor = {
+						enumerable: true,
+						configurable: true,
+						writable: true
+					};
```
devpaul, line 138: Would you add a test to observe a property that does not exist on the target?

src/observe.ts
```diff
@@ -0,0 +1,330 @@
...
+	protected _target: any;
+
+	constructor(kwArgs?: KwArgs) {
+		this._listener = kwArgs.listener;
+		this._propertyStore = {};
+		this._target = kwArgs.target;
+	}
+}
+
+export interface KwArgs {
```
devpaul, line 19: `KwArgs` and `ObserveArgs` appear to be identical. Can we remove one?

src/observe.ts
```diff
@@ -0,0 +1,330 @@
+import has from './has';
+import { Handle, Hash } from './interfaces';
+import { getPropertyDescriptor, isIdentical } from './lang';
+import { queueMicroTask } from './queue';
+import Scheduler from './Scheduler';
+
+export class BaseObjectObserver {
```
devpaul, line 7: Why are we exporting `BaseObjectObserver`? We don't appear to be using it anywhere else.

bryanforbes:
Closing for #80.

Issue #37, opened 2015-06-12
Implement `on.once` and `on.pausable`


kfranqueiro:
We've had a few discussions back and forth on this topic (largely connected to the discussion around extension events), and I did some [fumbling around myself](https://gist.github.com/kfranqueiro/b61a0d4d1939063842b2).

Ultimately, we have decided to implement `once` and `pausable` APIs largely as they existed in Dojo 1.x (though they would be implemented as separate exports of the `on` module in this case).
- They will accept the same arguments as `on`
- `once` will return an object implementing `Handle`
- `pausable` will return an object implementing a `PausableHandle` which should extend `Handle` with `pause()` and `resume()`

PR #43, opened 2015-06-17
once / pausable

bitpshr:
This PR includes implementations of `once` and `pausable` with basic tests as per #37.


src/on.ts
```diff
@@ -104,3 +104,58 @@ export default function on(target: any, type: any, listener: any, capture?: bool
 
 	throw new TypeError('Unknown event emitter object');
 }
+
+/**
+ * Provides a mechanism for listening to the next occurence of events from event
```
kfranqueiro: Typo (occurrence)

src/on.ts
```diff
@@ -104,3 +104,58 @@ export default function on(target: any, type: any, listener: any, capture?: bool
...
+ * emitters, Evented objects, or DOM nodes.
+ * @param target Target to listen for event on
+ * @param type Event type(s) to listen for; may be strings or extension events
```
kfranqueiro: I realize now that I neglected this before when I cleared out extension events, but we should remove mention of them from jsdoc comments.

src/on.ts
```diff
@@ -104,3 +104,58 @@ export default function on(target: any, type: any, listener: any, capture?: bool
...
+
+	let handle = <PausableHandle> on(target, type, function () {
+		if (!paused) {
+			return listener.apply(this, arguments);
+		}
+	}, capture);
+
+	handle.pause = function () {
+		paused = true;
+		return this;
```
kfranqueiro: What was the motivation for returning the handle again from pause/resume?

src/on.ts
```diff
@@ -104,3 +104,58 @@ export default function on(target: any, type: any, listener: any, capture?: bool
...
+	};
+
+	handle.resume = function () {
+		paused = false;
+		return this;
+	};
+
+	console.log(handle);
```
kfranqueiro: Oops, this shouldn't be here :P

src/on.ts
kitsonk: Good question... of course the interface specifies `PausableHandle`, but should it?  There is no functional value in handle chaining.  Should maybe just be `void`?

src/on.ts
```diff
@@ -104,3 +104,58 @@ export default function on(target: any, type: any, listener: any, capture?: bool
...
+ * @param listener Callback to handle the event when it fires
+ * @param capture Whether the listener should be registered in the capture phase (DOM events only)
+ * @return A handle which will remove the listener when destroy is called
+ */
+export function pausable(target: EventTarget, type: string | string[], listener: EventCallback, capture?: boolean): PausableHandle;
+export function pausable(target: EventEmitter | Evented, type: string | string[], listener: EventCallback): PausableHandle;
+export function pausable(target: any, type: any, listener: any, capture?: boolean): PausableHandle {
+	let paused: boolean;
+
+	let handle = <PausableHandle> on(target, type, function () {
```
kitsonk: `const handle`?

src/interfaces.d.ts
```diff
@@ -6,6 +6,11 @@ export interface Handle {
 	destroy(): void;
 }
 
+export interface PausableHandle extends Handle {
+	pause(): PausableHandle;
```
kitsonk: As the below, why is this `PausableHandle` and not `void`?

src/on.ts
bitpshr: I did this mainly for the sake of convenience, but I realize now that if you're calling `pause` or `resume` you already have the handle you need.

bitpshr:
This should be ready for another review @kfranqueiro and @kitsonk. Thanks in advance.

kfranqueiro:
I decided to move the `PausableHandle` interface into `on.ts` since it's rather specific to one API in there.  If we find another use for it we could move it into `interfaces.d.ts` as this PR originally proposed.

PR #46, opened 2015-06-19
AMD text loader plugin

bitpshr:
This PR includes an AMD text plugin and functional tests to maintain a clean environment between cases.


kfranqueiro:
This branch was accidentally created on top of another branch for another feature.  You should be able to do a `git rebase -i master` to delete the commit that doesn't belong on this one.

Also, see https://github.com/dojo/loader/pull/16 for previous discussion, particularly https://github.com/dojo/loader/pull/16#issuecomment-113639314 which indicates more work that could be potentially done here.

bitpshr:
@kfranqueiro @edhager I exposed the `get` API externally, which I agree could be useful. I added tests as well, and killed the erroneous commits that were in this history. Should be ready for another review.

package.json
```diff
@@ -23,6 +23,7 @@
 		"grunt-ts": "4.2.0-beta.1",
 		"grunt-tslint": "2.0.0",
 		"intern": "2.2.2",
+		"leadfoot": "1.5.1",
```
kitsonk, line 26: Why do we have to directly add this dependency?

src/text.ts
```diff
@@ -0,0 +1,114 @@
...
+else {
+	getText = function(): void {
+		throw new Error('dojo/text not supported on this platform');
+	};
+}
+
+/*
+ * Cache of previously-loaded text resources
+ */
+let textCache: { [key: string]: any; } = {};
```
kitsonk, line 56: `const`?

src/text.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/*
+ * Cache of pending text resources
+ */
+let pending: { [key: string]: any; } = {};
```
kitsonk, line 61: `const`?

kitsonk:
Can we get this rebased and the questions addressed?

PR #47, opened 2015-06-22
Event utility functions

bitpshr:
This PR includes implementations of `debounce`, `throttle`, and `throttleAfter`, as well as `createTimer` and `createInterval`, all of which return a `Handle`. 90%+ test coverage is also provided. For now, I stuck these in a top-level `utility.ts` file. I'm not sure if we should make this specific to events naming-wise just yet.


kfranqueiro:
Feedback, by way of changes I'm incorporating into the final commit:
- Added jsdoc comments
- Removed createInterval since it wasn't being used internally and we tend to encourage setTimeout over setInterval due to potential collapsing/skipping of intervals with the latter
- Removed usage of createTimer in debounce in favor of setTimeout, since these functions tend to be inherently performance-sensitive
- Changed delay in createTimer from default 0 to optional, since setTimeout already handles undefined delay.  This also brings the tests to 100% coverage (the only thing that wasn't covered was not setting delay)
- delay is now required for debounce/throttle/throttleAfter, since defaulting to 0 seems to mostly defeat the point of the functions to begin with, and I'm not sure we should have an opinion on a default
- Removed unused `has` import
- Replaced `equal` with `strictEqual` in assertions
- Added descriptions to many assertions
- Added assertions to throttle and throttleAfter to assert which end of the interval the original function is invoked on
- Removed an unnecessary setTimeout step in one of the throttle tests, since throttle calls the function immediately on the first call in each interval
- Removed unnecessary quotes around single-word test function names
- Fixed trailing comma at the end of the tests module
- Fixed indentation of test module
- Removed trailing whitespace

kfranqueiro:
I am keeping this code where it lives in `util` for now, though I imagine we might form a more concrete opinion of where it should ultimately live if we find more utilities to add.

Issue #34, opened 2015-06-12
Improve request tests


bryanforbes:
The request tests need to be improved in the following ways:
- [x] Stop shimming XMLHttpRequest
- [x] Implement an echo server:
  - [x] Proxies to Intern's server for everything except `/__echo/`
  - [x] When `/__echo/` is requested, parse the query string, headers, and (for POST and PUT) the data and echo them back in a JSON response (something like `{ query: { ... }, headers: { ... }, data: { ... } }`)
  - [x] If the `responseType` query parameter is included, send data back in the appropriate format (for instance, `responseType=document` should return HTML)
- [x] Change the intern configuration to hit the echo server
- [x] Write tests to improve coverage, including tests for different response types

kfranqueiro:
@bryanforbes, should stuff in [dojo's tests folder](https://github.com/dojo/dojo/tree/master/tests) (e.g. the stuff under services and its setup in `dojo.intern.js`) be useful for reference for this, or does this involve a different approach?  I'm a little confused since it seems as if the service there is only set up if you're running intern-runner, not client, so does that mean that Dojo 1's request intern tests only work via runner?

I'd like to make it as clear as possible how to get started on this, given that this task is being delegated.

bryanforbes:
Sorry that I wasn't more specific. Yes, this should follow a similar approach to what Dojo 1 does, except that implementing a bunch of services, this should only implement one service (`__echo`). The relative parts from Dojo 1 are:
- [This section](https://github.com/dojo/dojo/blob/master/tests/dojo.intern.js#L14-L28) of `dojo.intern.js` for starting the services and for configuring the proxy
- [services/main.js](https://github.com/dojo/dojo/blob/master/tests/services/main.js) to see how we're writing the server. Keep in mind that this is using `jsgi-node` for the server which is substantially different than connect or express. I don't care what the server is written in, but it'll have to handle multipart data and proxying.

bryanforbes:
Also, using this approach, there will be two ways to run the unit tests: using the Intern runner, or starting the Intern runner with the `proxyOnly` setting and then hit the server manually (the easiest way would be to hit `run.html`).

PR #48, opened 2015-07-07
Improve request tests.

maier49:
Refs #34

Now provides an echo server on port 9001 that proxies most
calls to the intern server, except for requests to /__echo/.

Calls to /__echo/ return a JSON response that contains the
parsed query string, headers, and payload from the request.

If a responseType query parameter is specified, a response
in the format requested will be returned if available.
Currently, 'html', 'document', 'xml', 'json', and 'gif'
are acceptable values for responseType. If any format other
than JSON is requested the response body will not contain
the payload, query or headers from the request.

The X-Requested-With header is no longer being used for
testing custom headers because IE 11 was modifying the
value before sending to the server.

Added tests for different content types and to improve
coverage.


maier49:
Currently ~~two~~ three of these tests fail because request/xhr is not handling the query property from XhrRequestOptions. Once that is fixed those tests should start passing.

kfranqueiro:
We're reassigning this to @maier49 so that he can test #49 when addressing that too.

maier49:
This now includes @mwistrand 's code for handling the query and cacheBust options, as well as test cases for those changes, so it should resolve #34 and #49.

src/request/util.ts
```diff
@@ -0,0 +1,24 @@
...
+	if (typeof query === 'object') {
+		query = new UrlSearchParams(query).toString();
+	}
+
+	if (options.cacheBust) {
+		const cacheBust = String(Date.now());
+		query += query ? '&' + cacheBust : cacheBust;
+	}
+
+	const separator = url.indexOf('?') > -1 ? '&' : '?';
```
bryanforbes, line 18: Is there a reason we're doing all of this logic here and on line 19 when we have it in `UrlSearchParams`?

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
-			xhrRequest(getRequestUrl('foo.json'), { timeout: 1 })
+		'timeout'() {
+			if (!echoServerAvailable) {
+				this.skip('No echo server available');
+			}
+			return xhrRequest(getRequestUrl('foo.json'), { timeout: 1 })
 				.then(
-					dfd.resolve.bind(dfd),
-					dfd.callback(function (error: Error): void {
+					null,
```
bryanforbes: This should probably throw or have an assertion that will always fail so it doesn't have to time out.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
+					}
 				);
+		},
+
+		'.get with URL query'() {
+			if (!echoServerAvailable) {
+				this.skip('No echo server available');
+			}
+			return xhrRequest('/__echo/xhr?color=blue&numbers=one&numbers=two', {
+				handleAs: 'json'
```
bryanforbes, line 70: `responseType` and then you don't have to parse down below

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
+				if (!echoServerAvailable) {
+					this.skip('No echo server available');
+				}
+				return xhrRequest('/__echo/xhr?color=blue&numbers=one&numbers=two', {
+					query: new UrlSearchParams({
+						foo: [ 'bar', 'baz' ],
+						thud: 'thonk',
+						xyzzy: '3'
+					}).toString(),
+					handleAs: 'json'
```
bryanforbes, line 155: `responseType` and then you don't have to parse below

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
+				if (!echoServerAvailable) {
+					this.skip('No echo server available');
+				}
+				return xhrRequest('/__echo/xhr', {
+					query: new UrlSearchParams({
+						foo: [ 'bar', 'baz' ],
+						thud: 'thonk',
+						xyzzy: '3'
+					}).toString(),
+					handleAs: 'json'
```
bryanforbes, line 182: `responseType` and then you don't have to parse below

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
+				if (!echoServerAvailable) {
+					this.skip('No echo server available');
+				}
+				return xhrRequest('/__echo/xhr', {
+					query: {
+						foo: [ 'bar', 'baz' ],
+						thud: 'thonk',
+						xyzzy: '3'
+					},
+					handleAs: 'json'
```
bryanforbes, line 207: `responseType` and then you don't have to parse below

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,497 @@ import * as registerSuite from 'intern!object';
...
+						xyzzy: '3'
+					}).toString(),
+					handleAs: 'json'
+				}).then(
+					function (response: any) {
+						const query = JSON.parse(response.data).query;
+						assert.strictEqual(query.color, 'blue');
+						assert.deepEqual(query.numbers, [ 'one', 'two' ]);
+						assert.strictEqual(query.thud, 'thonk');
+						assert.strictEqual(query.foo.length, 2);
```
bryanforbes: Can the check for `query` be done in a single `assert.deepEqual`?

src/request/util.ts
maier49, line 18: I removed the redundant logic at the beginning of the function, but I wasn't sure how line 19 could also be removed, since UrlSearchParams doesn't deal with the presence or absence of '?'.

tests/unit/request/xhr.ts
maier49, line 155: 'json' is explicitly excluded from the response types handled by xhr, unless I'm missing something.

tests/unit/request/xhr.ts
maier49: Should it throw a RequestTimeoutError? If it throws something else the test fails because the error.name doesn't match in the error handler. But if we're throwing that explicitly what would this really be testing?

tests/unit/request/xhr.ts
bryanforbes, line 155: https://github.com/dojo/core/blob/e962a14fbaf2ea946f69d49a4d59811f2cdaecdc/src/request.ts#L194

tests/unit/request/xhr.ts
maier49: Updated to use `assert.deepEqual` on the whole object.

tests/unit/request/xhr.ts
bryanforbes: Sorry, I didn't explain myself very well. The code should probably do this:

``` ts
return xhrRequest(getRequestUrl('foo.json'), { timeout: 1 })
    .then(
        function () {
            assert(false, 'Request should have timed out');
        },
        function (error: Error) {
```

src/request/util.ts
bryanforbes, line 18: Currently, `URLSearchParams` will take a full URL and return a full URL. This is the wrong behavior according to the spec (which I just checked), so I thought we could use that to our advantage. Leave this line. Nice work on removing the redundant logic :).

bryanforbes, line 18: Aaaaaaaaaand ignore me on `UrlSearchParams` being wrong. I'm on a roll today.

tests/unit/request/xhr.ts
maier49: I added this, and the test immediately started failing. It is resolving successfully, but with a statusCode of 0 whenever it timed out. Despite being an invalid HTTP status, 0 is a valid value for statusCode in an XMLHttpRequest, and indicates an error of some sort. But, I'm not sure why it is consistently firing when the request times out. 

Wrapping the resolution of the request in src/request/xhr.ts with a check to make sure the status code is between 200 and 400 and rejecting otherwise creates the appropriate behavior in the sense that the request fails, but still results in the wrong error type because reject gets fired before the timeout handler.

tests/typings/tsd.d.ts
```diff
@@ -1,3 +1,6 @@
 /// <reference path="intern/intern.d.ts" />
 /// <reference path="benchmark/benchmark.d.ts" />
 /// <reference path="sinon/sinon.d.ts" />
+/// <reference path="formidable/formidable.d.ts" />
+/// <reference path="http-proxy/http-proxy.d.ts" />
+/// <reference path="services/echo.d.ts" />
```
jason0x43: Are we still doing typings in a tsd, or should they just be defined in the `tsconfig.json`?

maier49: Good catch, I removed the tsd file and everything is still working.

src/request/xhr.ts
```diff
@@ -67,8 +69,15 @@ export default function xhr<T>(url: string, options: XhrRequestOptions = {}): Re
 
 				response.statusCode = request.status;
 				response.statusText = request.statusText;
-
-				resolve(response);
+				if (response.statusCode >= 200 && response.statusCode < 400) {
```
jason0x43, line 72: Would this be better as just `response.statusCode < 400` since the 100 codes aren't errors?

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'), { method: 'get' })
+		specified() {
+			if (!echoServerAvailable) {
+				this.skip('No echo server available');
+			}
+			return xhrRequest(getRequestUrl('foo.json'), { method: 'get' })
 				.then(
-					dfd.callback(function (response: any): void {
+					function (response: any) {
```
jason0x43: Single callbacks generally go on the same line as `then(`

tests/intern.ts
```diff
@@ -1,9 +1,17 @@
 import intern = require('intern');
+import has = require('intern/dojo/has');
+import topic = require('intern/dojo/topic');
+import echo = require('intern/dojo/has!host-node?./services/echo');
 
+if (has('host-node') && intern.mode === 'runner') {
+	echo.start().then(function (server: any) {
+		topic.subscribe('runner/end', server.close);
```
jason0x43: The topic needs an initial forward slash ("/runner/end"). Also, `server.close` should be called in a callback or bound.

tests/unit/request/xhr.ts
jason0x43, line 70: The xhr request module doesn't do that kind of processing.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-					dfd.reject.bind(dfd)
+					}
 				);
 		},
 
-		'default'(): void {
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'))
+		'default'() {
+			if (!echoServerAvailable) {
```
jason0x43, line 56: This test isn't using the echo server.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
+			);
+		});
+	},
+
 	'HTTP methods': {
-		specified(): void {
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'), { method: 'get' })
+		specified() {
+			if (!echoServerAvailable) {
```
jason0x43, line 46: This test isn't using the echo server

tests/services/echo.ts
```diff
@@ -0,0 +1,176 @@
...
+		else {
+			proxy.proxyRequest(request, response, {
+				host: 'localhost',
+				port: 9000
+			});
+		}
+	});
+
+	const promise: Promise<http.Server> = new Promise(function (resolve, reject) {
+		server.on('listening', resolve);
```
jason0x43: Pass the server to `resolve` so the shutdown code in `intern.ts` will work.

tests/unit/request/xhr.ts
jason0x43, line 155: That's not xhr

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
+			}
+			return xhrRequest('/__echo/xhr', { timeout: 1 })
 				.then(
-					dfd.resolve.bind(dfd),
-					dfd.callback(function (error: Error): void {
+					function () {
+						assert(false, "Should have timed out");
+					},
+					function (error: Error) {
 						assert.strictEqual(error.name, 'RequestTimeoutError');
```
jason0x43, line 114: At least in Chrome, this error has name "Error", so this test fails.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-					})
+					}
 				);
 		},
 
-		'user and password'(): void {
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'), {
+		'user and password'() {
+			if (!echoServerAvailable) {
```
jason0x43, line 120: This test isn't using the echo server.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-				dfd.reject.bind(dfd)
+				}
 			);
 		},
 
-		'auth, without user or password'(): void {
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'), {
+		'auth, without user or password'() {
+			if (!echoServerAvailable) {
```
jason0x43, line 133: This test isn't using the echo server.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
+				}
+				return xhrRequest('/__echo/normalize', {
+					headers: {
+						'CONTENT-TYPE': 'arbitrary-value'
+					}
+				}).then(
+					function (response: any) {
+						const data = JSON.parse(response.data);
+						assert.isUndefined(data.headers['CONTENT-TYPE']);
+						assert.include(data.headers['content-type'], 'arbitrary-value');
```
jason0x43: Why `include` rather than `strictEqual`? There's also the more specific `propertyVal`.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-		'.getHeader'(): void {
-			const dfd = this.async();
-			xhrRequest(getRequestUrl('foo.json'))
-				.then(
-					dfd.callback(function (response: any): void {
-						const length: number = Number(response.getHeader('content-length'));
-						assert.strictEqual(length, response.data.length);
-					}),
-					dfd.reject.bind(dfd)
+			'default'() {
```
jason0x43: Is this test adding anything that wasn't covered by other tests?

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-				.then(
-					dfd.callback(function (response: any): void {
-						const foo: string = response.data.getElementsByTagName('foo')[0].getAttribute('value');
-						assert.strictEqual(foo, 'bar');
-					}),
-					dfd.reject.bind(dfd)
-				);
+	'response object': {
+		properties() {
+			if (!echoServerAvailable) {
```
jason0x43, line 432: This test doesn't use the echo server.

tests/unit/request/xhr.ts
```diff
@@ -2,168 +2,520 @@ import * as registerSuite from 'intern!object';
...
-			xhrRequest(getRequestUrl('foo.json'))
-				.then(
-					dfd.callback(function (response: any): void {
-						const data: {} = JSON.parse(response.data);
-						assert.deepEqual(data, { foo: 'bar' });
-					}),
-					dfd.reject.bind(dfd)
-				);
+		'.getHeader'() {
+			if (!echoServerAvailable) {
```
jason0x43, line 445: This test doesn't use the echo server.

src/request/xhr.ts
maier49, line 72: This ties into your comment on the 'timeout' test that's failing. When the request times out, prior to the ontimeout callback getting fired, this code gets fired with a status code of 0. I looked into it and apparently 0 is  a valid value for XMLHttpRequest's statusCode property, indicating an error. I could change this to `response.statusCode < 400 && response.statusCode !== 0`, but I'm still not sure why this is getting called with a statusCode of 0, and how we want to handle that with regards to the 'timeout' test.

maier49:
With the exception of the issue regarding status codes and the timeout test I mentioned in a comment above, all the issues @jason0x43 commented on should be resolved now.

tests/services/echo.ts
```diff
@@ -0,0 +1,178 @@
...
+	const multipartRE = /^multipart\/form-data;/;
+
+	return function (request: any, response: any) {
+		const headers = request.headers;
+		if (headers['content-type'] && multipartRE.test(headers['content-type'])) {
+			request.data = new Promise(function (resolve, reject) {
+				const parser = new formidable.IncomingForm();
+				parser.parse(request, function (err: any, fields: formidable.Fields, files: formidable.Files) {
+					if (err) {
+						reject(err);
```
jdonaghue, line 24: this should probably return, currently it will flow through

maier49, line 24: Added a return after reject.

PR #53, opened 2015-07-15
simplify the implementation of 'isObject', fixes #45 refs #45

### LINKED ISSUE #45: Concerns regarding lang's internal isObject function
This function is used to determine when something should be copied/assigned recursively.  It currently returns true for things that report `typeof` as `object` that are not Arrays or RegExps.  But I would think there are other things (e.g. Date) that should also return false here.

@mwistrand suggested `Object.prototype.toString.call(item) === '[object Object]'` as a possible alternative that would avoid this slippery slope.  Can anyone think of a reason that would be undesirable?

jdonaghue:


kitsonk:
:+1:

PR #55, opened 2015-07-16
add tests for aspect#on, enhance test coverage, fixes #52, refs #52

### LINKED ISSUE #52: Missing tests for aspect.on
In dojo-core, `aspect.after` has been split to `aspect.after` and `aspect.on` (the latter serving the same purpose that `aspect.after` with `receiveArguments` set to `true` served in Dojo 1).

However, we're currently not testing `aspect.on` at all.  I expect we probably only need a couple of basic tests to hit 100% coverage, since most of the ensuing code path is the same as `aspect.after`.

jdonaghue:


kitsonk:
:+1:

src/has.ts
```diff
@@ -3,6 +3,60 @@ import { Hash } from './interfaces';
 export const cache: Hash<any> = Object.create(null);
 const testFunctions: Hash<() => any> = Object.create(null);
 
+export const dynamic: boolean = true;
```
bryanforbes: Why are these changes included in the pull request?

jdonaghue: Shoot I must have used the same branch

tests/unit/aspect.ts
```diff
@@ -171,6 +171,41 @@ registerSuite({
...
 		},
 
+		'.on()': {
+			'advising function returns undefined, returns original result'() {
+				const aspectStub1 = sinon.stub();
+
+				aspect.on(obj, 'method', aspectStub1);
+
+				assert.strictEqual(obj.method(0), 1);
+				assert.isTrue(aspectStub1.calledAfter(methodSpy));
```
bryanforbes: There should probably be checks for what `aspectStub1` was called with and the number of times it was called.

tests/unit/aspect.ts
```diff
@@ -171,6 +171,41 @@ registerSuite({
...
+			},
+
+			'advising function returns defined values, returns advising function result'() {
+				const aspectStub1 = sinon.stub().returns(2);
+
+				aspect.on(obj, 'method', aspectStub1);
+
+				assert.strictEqual(obj.method(0), 2);
```
bryanforbes, line 193: I know that this was the behavior in Dojo 1, but do we really want this same behavior in Dojo 2? cc @kfranqueiro

tests/unit/aspect.ts
```diff
@@ -171,6 +171,41 @@ registerSuite({
...
+				const aspectStub3 = sinon.stub().returns(6);
+
+				aspect.on(obj, 'method', aspectStub1);
+				aspect.on(obj, 'method', aspectStub2);
+				aspect.on(obj, 'method', aspectStub3);
+
+				assert.strictEqual(obj.method(0), 6);
+				assert.isTrue(aspectStub1.calledAfter(methodSpy));
+				assert.isTrue(aspectStub2.calledAfter(methodSpy));
+				assert.isTrue(aspectStub3.calledAfter(methodSpy));
```
bryanforbes: There need to be more checks: that the later stubs run after the earlier ones, that they are all called with the correct arguments, and that they were called the correct number of times.

tests/unit/has.ts
```diff
@@ -96,6 +96,79 @@ registerSuite({
 				hasAdd('one', true, true);
 				assert.isTrue(has('one'));
 			}
+		},
```
bryanforbes: These are also included and shouldn't be.

bryanforbes:
I reviewed the PR. Please fix the issues I highlighted. Thanks!

tests/unit/aspect.ts
kfranqueiro, line 193: We're talking about `aspect.on`, so I'd potentially expect the return value to be meaningless.  It currently inherits `after` + `receiveArguments` logic from Dojo 1 so it can still influence the return value.

Do people often want to influence the return value when `receiveArguments` is true?  Usually it's more useful to have the returned value for processing when you actually want to change it (i.e. `receiveArguments` false / `aspect.after` in Dojo 2).

Another thing that is potentially awkward is that I would figure that right now, `after` and `on` are sharing the same queue.

I think these are things to consider but shouldn't block this PR.  But we may want to leave this behavior (of whether it modifies the return value) as undefined for now.

bryanforbes, line 193: IMO, if someone wants to modify the return value, they should be using `aspect.after` and `aspect.on` should be simply about notification of the method running (like an event).

kitsonk, line 193: > IMO, if someone wants to modify the return value, they should be using aspect.after and aspect.on should be simply about notification of the method running (like an event).

In fact, I don't like that it implies that it resolve synchronously too.  I would almost want to remove this feature to keep people from doing stupid stuff.  I would almost be tempted at this point to make `.on` resolve as a scheduled microtask at the end of the turn and @kfranqueiro has an interesting point... someone coming along doing an `on` before an `after` would have a rather surprising unintuitive outcome, especially if the main way of modifying method behaviour will be aspect advice.

While this PR reflects current functionality, I think we should consider the overall functionality now before we regret it...

kitsonk, line 193: Ok, I have clearly missed something... there is no `aspect#on` in Dojo 1.  Why do we need `core/aspect#on` in Dojo 2 when we also have `core/on`?  Won't that lead to a significant amount of confusion (like it just did for me?)  Which do I use, when do I use it, why doesn't it work?  It will be `dojo.byId` and `dijit.byId` all over again!

I took a look at other AOP implementations and I don't see any `on` advice.  I have seen only `before`, `after`, and `around` with some variations like "after throwing" and "after finally".  `aspect#on` might imply (thought it doesn't technically do this at the moment) "after finally" so wouldn't it be better to call it `afterFinally`?

kfranqueiro, line 193: As I alluded to in my previous comment, `aspect.on` is equivalent to Dojo 1's `aspect.after(obj, method, func, true)`, while `aspect.after` here is the same as Dojo 1's `aspect.after(obj, method, func, false)`.

IIRC @bryanforbes came up with the idea of separating its API surface, since it really is more akin to an event handler (in that it passes the original arguments along to the advising function, not the return value) than to AOP.

I believe cujojs/meld has the [same idea](https://github.com/cujojs/meld/blob/master/docs/api.md#meldon).

Your concern over confusion over which `on` to use is valid.  At a technical level it comes down to differentiating event emitters and events from generic objects using methods as join points.  To a degree I'd expect the same confusion to already be possible in Dojo 1 anyway (especially given widgets' legacy behavior of events being registered via AOP).

bryanforbes, line 193: @kitsonk Can you link to the AOP implementations you looked at?

kitsonk, line 193: - [Spring AOP](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/aop.html)
- [AspectJ](https://eclipse.org/aspectj/doc/next/progguide/semantics-advice.html)
- [Go AOP](http://go.aopphp.com/docs/pointcuts-and-advices/) this one doesn't have _After Finally_ advice

I might be interesting to deliver some of these other types of advice (_After Finally_ and _After Throwing_).

PR #56, opened 2015-07-18
Use native `Object.assign` when available. Fixes #54.

### LINKED ISSUE #54: Shim Object.assign?
Instead of creating the core/lang::assign as it is now, wouldn't it be better to shim so it offloads to native assign (and keeps it functionally aligned?)... something like:

``` typescript
interface ObjectConstructor {
    assign(target: any, ...sources: any[]): any;
}

/* Adapted from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Polyfill */
let assign = Object.assign ? Object.assign : function(target: any, ...sources: any[]): any {
    'use strict';
    if (typeof target === 'undefined' || target === null) {
        throw new TypeError('Cannot convert first argument to object');
    }

    const to = Object(target);
    for (let i = 0; i < sources.length; i++) {
        let nextSource = sources[i];
        if (typeof nextSource === 'undefined' || nextSource === null) {
            continue;
        }
        nextSource = Object(nextSource);

        const keysArray = Object.keys(Object(nextSource));
        for (let nextIndex = 0; nextIndex < keysArray.length; nextIndex++) {
            const nextKey = keysArray[nextIndex];
            const desc = Object.getOwnPropertyDescriptor(nextSource, nextKey);
            if (typeof desc !== 'undefined' && desc.enumerable) {
                to[nextKey] = nextSource[nextKey];
            }
        }
    }
    return to;
};
```

mwistrand:
- `lang.assign` delegates to `Object.assign` when avaialble.
- Implement `has('object-assign')` test.

### DIFF
--- a/src/has.ts
+++ b/src/has.ts
@@ -64,6 +64,7 @@ add('dom-mutationobserver', function(): boolean {
 add('microtasks', function () {
 	return has('promise') || has('host-node') || has('dom-mutationobserver');
 });
+add('object-assign', typeof (<any> Object).assign === 'function');
 add('object-observe', typeof (<any> Object).observe === 'function');
 add('postmessage', typeof postMessage === 'function');
 add('promise', typeof global.Promise !== 'undefined');
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -34,6 +34,10 @@ interface MixinArgs {
 	target: {};
 }
 
+interface ObjectAssignConstructor extends ObjectConstructor {
+	assign(target: {}, ...sources: {}[]): {}
+}
+
 function _mixin(kwArgs: MixinArgs): {} {
 	const deep = kwArgs.deep;
 	const inherited = kwArgs.inherited;
@@ -73,10 +77,10 @@ function _mixin(kwArgs: MixinArgs): {} {
  * @param sources Any number of objects whose enumerable own properties will be copied to the target object
  * @return The modified target object
  */
-export let assign = has('object-assign') ?
-	(<any> Object).assign :
-	function assign(target: {}, ...sources: {}[]): {} {
-		return (typeof (<any> Object).assign === 'function') ? (<any> Object).assign.apply(null, arguments) : _mixin({
+export const assign = has('object-assign') ?
+	(<ObjectAssignConstructor> Object).assign :
+	function (target: {}, ...sources: {}[]): {} {
+		return _mixin({
 			deep: false,
 			inherited: false,
 			sources: sources,
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -73,14 +73,16 @@ function _mixin(kwArgs: MixinArgs): {} {
  * @param sources Any number of objects whose enumerable own properties will be copied to the target object
  * @return The modified target object
  */
-export function assign(target: {}, ...sources: {}[]): {} {
-	return _mixin({
-		deep: false,
-		inherited: false,
-		sources: sources,
-		target: target
-	});
-}
+export let assign = has('object-assign') ?
+	(<any> Object).assign :
+	function assign(target: {}, ...sources: {}[]): {} {
+		return (typeof (<any> Object).assign === 'function') ? (<any> Object).assign.apply(null, arguments) : _mixin({
+			deep: false,
+			inherited: false,
+			sources: sources,
+			target: target
+		});
+	}
 
 /**
  * Creates a new object from the given prototype, and copies all enumerable own properties of one or more


src/lang.ts
```diff
@@ -73,14 +73,16 @@ function _mixin(kwArgs: MixinArgs): {} {
...
  */
-export function assign(target: {}, ...sources: {}[]): {} {
-	return _mixin({
-		deep: false,
-		inherited: false,
-		sources: sources,
-		target: target
-	});
-}
+export let assign = has('object-assign') ?
```
kitsonk: An alternative way of doing this, instead of the `any` cast (that allows the interface to be preserved too) is the following:

``` typescript
interface ObjectAssignConstructor extends ObjectConstructor {
    assign(target: {}; ...sources: {}[]): {};
}

function assign(target: {}; ...sources: {}[]): {} {
    return (typeof (<ObjectAssignConstructor> Object).assign === 'function') ? (<ObjectAssignConstructor> Object).assign.apply(null, arguments) : _mixin({
        deep: false,
        inherited: false,
        sources: sources,
        target: target
    });
```

src/lang.ts
```diff
@@ -73,14 +73,16 @@ function _mixin(kwArgs: MixinArgs): {} {
...
+	(<any> Object).assign :
+	function assign(target: {}, ...sources: {}[]): {} {
+		return (typeof (<any> Object).assign === 'function') ? (<any> Object).assign.apply(null, arguments) : _mixin({
+			deep: false,
```
kitsonk, line 84: Just realised you are checking the has flag in line 76 and then executing code that should never be accessible because this branch shouldn't be followed if `has('object-assign')` is true.

src/has.ts
```diff
@@ -64,6 +64,7 @@ add('dom-mutationobserver', function(): boolean {
 add('microtasks', function () {
 	return has('promise') || has('host-node') || has('dom-mutationobserver');
 });
+add('object-assign', typeof (<any> Object).assign === 'function');
```
kitsonk, line 67: We should solidify the `has` naming...  I know somewhere else we discussed that we would prefer to use `es6-object-assign` for something like this (and `object-observe` should be [in theory] `es7-object-observe` and `promise` should be `es6-promise`).

kitsonk:
@mwistrand the problem you are running into is what I ran into on Microsoft/TypeScript#3889.  It isn't static methods, it is that you are trying to extend it within the module and you cannot access the global namespace for interfaces from within a module.

src/lang.ts
kfranqueiro, line 84: Following that logic, wouldn't your `ObjectAssignConstructor` interface then be needed both in the `has` test and here?

I'm usually pro-avoiding-any, but it seems like a whole lot of pointless runaround for something that ultimately doesn't matter in this case (at least as far as type information for the external API is concerned).

Edit: wait, I guess it does matter, if we're assigning Object.assign directly in that case.  Darnit.

kitsonk, line 84: Yes, I think we have to have it where we assign `Object.assign` to the function to preserve the signature... Not wholly necessary in the `has` check though.

kfranqueiro, line 84: Sounds fair.

src/lang.ts
kitsonk: Oh, also, `export const` right?

src/lang.ts
mwistrand, line 84: Excellent! Another mistake to keep me humble! Regarding the extended type, that proposal does seem like the best way to ensure the interface is preserved.

src/has.ts
mwistrand, line 67: I vaguely remember that discussion, but also vaguely remember no definitive consensus, so I decided to follow the convention already in place.

mwistrand:
@kitsonk Thanks for passing along that TS issue. I was partly hoping that built-in interfaces could be opened up for extension (like modules/classes in Ruby), but that's clearly not the case. I'm not convinced, though, that `declare var Object: ObjectAssignConstructor` is more elegant than casting `Object`, so I stuck with your initial suggestion.

src/has.ts
kitsonk, line 67: Yeah, it wasn't specifically directed at you, but more the wider bit reminding that we didn't close it off... I guess we can go back later and clean that up.

kitsonk:
I think they will be, once Microsoft/TypeScript#983 is solved.  You can extend them if you aren't in a module.  The other solution is to add them to the projects .d.ts, but then you are "changing" the global `Object` for that project, which you might not actually want to do.  So the "limitation" in this case actually benefits I think for feature detection.

src/has.ts
kitsonk, line 67: I have opened an issue in dojo/guidelines#5.

PR #57, opened 2015-07-23
changes to has module to facilitate loading it as a plugin via loader

jdonaghue:


src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+ * Resolves resourceId into a module id based on possibly-nested tenary expression that branches on has feature test value(s).
+ *
+ * @param resourceId The id of the module
+ * @param toAbsMid Resolves a relative module id into an absolute module id
+ */
+export function normalize(resourceId: string, normalize: (moduleId: string) => string): string {
+	var tokens = resourceId.match(/[\?:]|[^:\?]*/g);
+	var i = 0;
+	function get(skip?: boolean): string {
+		var term = tokens[i++];
```
kitsonk: `const`?

src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+
+/**
+ * Resolves resourceId into a module id based on possibly-nested tenary expression that branches on has feature test value(s).
+ *
+ * @param resourceId The id of the module
+ * @param toAbsMid Resolves a relative module id into an absolute module id
+ */
+export function normalize(resourceId: string, normalize: (moduleId: string) => string): string {
+	var tokens = resourceId.match(/[\?:]|[^:\?]*/g);
+	var i = 0;
```
kitsonk: `let`?

src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+}
+
+/**
+ * Resolves resourceId into a module id based on possibly-nested tenary expression that branches on has feature test value(s).
+ *
+ * @param resourceId The id of the module
+ * @param toAbsMid Resolves a relative module id into an absolute module id
+ */
+export function normalize(resourceId: string, normalize: (moduleId: string) => string): string {
+	var tokens = resourceId.match(/[\?:]|[^:\?]*/g);
```
kitsonk: `const`?

src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+	else {
+		loaded();
+	}
+}
+
+/**
+ * Resolves resourceId into a module id based on possibly-nested tenary expression that branches on has feature test value(s).
+ *
+ * @param resourceId The id of the module
+ * @param toAbsMid Resolves a relative module id into an absolute module id
```
devpaul: parameter name should be `normalize`

src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+ */
+export function normalize(resourceId: string, normalize: (moduleId: string) => string): string {
```
kitsonk, line 28: Seems like normalize type should be an interface somewhere that is part of a `d.ts`?

tests/unit/has.ts
```diff
@@ -96,6 +96,79 @@ registerSuite({
 				hasAdd('one', true, true);
 				assert.isTrue(has('one'));
 			}
+		},
+
+		'has loader tests': {
```
devpaul, line 102: I haven't been able to get these tests passing using `grunt test`. I'm seeing errors like:

```
Error: Attempt to require unloaded module src/has!intern!object
  at contextRequire  <loader.ts:385:12>
  at module.require.result  <loader.ts:416:11>
  at Test.registerSuite.has loader tests.both feature and no-feature modules provided [as test]  <_build/tests/unit/has.js:100:35>
```

This is probably because the single argument require() expects the module to already be loaded.

I'm not certain we want to conflate testing has.normalize with require. Why don't we just bring in `normalize` and `load` and test these methods directly.

```
import { normalize as hasNormalize, load as hasLoad } from 'src/has';
```

This will verify the unit under test without testing the loader's require statement :)

```
'test'() {
    var expected = has('host-browser') ? 'intern' : 'intern!object';
    var actual = hasNormalize('host-browser?intern:intern!object', (<any> require).toAbsMid);
    assert.strictEqual(actual, expected);
},
```

tests/unit/has.ts
```diff
@@ -96,6 +96,79 @@ registerSuite({
...
+
+				hasAdd('abc', true);
+				hasAdd('def', false);
+
+				var result1 = require('src/has!abc?intern:intern!object');
+				assert.strictEqual(result1, featureModule);
+
+				var result2 = require('src/has!def?intern:intern!object');
+				assert.strictEqual(result2, noFeatureModule);
+			}
```
devpaul, line 202: Would you add a test for a chained ternary: e.g.

``` javascript
has.add('true', true);
has.add('false', false);
normalize('true?false?one:two:three')
normalize('true?true?one:two:three')
normalize('false?true?one:two:three')
```

jdonaghue:
thanks for the comments and feedback @kitsonk and @devpaul, I have made adjustments based on that in https://github.com/jdonaghue/core/commit/ca535666879ad1f9f602f344c08cc6f8fa80e6ec.

kitsonk:
See my thoughts in issue #59.  It would be good to rely upon an interface here.

jdonaghue:
That is a good idea. Is it sufficient to add the interface within this file and possibly export it? Or are you thinking it should go in `interfaces.d.ts`? Or somewhere else?

kfranqueiro:
I added some comments to #59.

src/has.ts
```diff
@@ -4,6 +4,58 @@ export const cache: Hash<any> = Object.create(null);
...
+ * @param resourceId The id of the module
+ * @param normalize Resolves a relative module id into an absolute module id
+ */
+export function normalize(resourceId: string, normalize: (moduleId: string) => string): string {
+	const tokens = resourceId.match(/[\?:]|[^:\?]*/g);
+	let i = 0;
+	function get(skip?: boolean): string {
+		const term = tokens[i++];
+		if (term === ':') {
+			// empty string module name, resolves to 0
```
baw, line 34: Should this say "resolves to null" instead?

tomdye:
Closing in favour of https://github.com/dojo/core/pull/122

PR #58, opened 2015-07-25
load: use Require interface from dojo-loader for AMDRequire

msssk:
The load module was providing an incomplete definition of the `Require` (`AMDRequire`) interface, and duplicating code from `dojo-loader`.

This commit updates `load.ts` to instead use the `Require` interface defined in `dojo-loader/loader`.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -1,3 +1,4 @@
+import { Require as AMDRequire } from 'dojo-loader/loader';
 import Promise from './Promise';
 
 declare var define: {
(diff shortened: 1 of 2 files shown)


msssk:
Hmmm... I'm not sure this is quite complete. The generated `dojo-core-2.0.0-pre.d.ts` (from `grunt dist`) doesn't know about the `dojo-loader` package - how do we fix that?

kitsonk:
I will try to look a bit later, but usually it is because you have exported something that dts-generator needs.  See SitePen/dts-generator#21.

kitsonk:
Hmmm actually it looks like maybe dst-generator doesn't handle export as...  Maybe open an issue on dts-generator?

kitsonk:
I think with TypeScript 1.6, we just need to update our `tsconfig.json` to reference the loaders typings now.  If possible, it would be create to get this PR updated, otherwise I can just land it by doing the changes off the current master.

bitpshr:
I reverified this; the [resulting typings](https://gist.github.com/bitpshr/07d79c64ede50b9c8f74) look correct @kitsonk and @msssk.

PR #65, opened 2015-08-21
Capture equals signs in URL query parameters

maier49:
Everything after the first equal sign in a query parameter
should be included in the value.


kfranqueiro:
This is pursuant to Section 5.1.6.2 of the URL spec: https://url.spec.whatwg.org/#urlencoded-parsing

> If _bytes_ contains a `=`, then let _name_ be the bytes from the start of _bytes_ up to but excluding its first `=`, and let _value_ be the bytes, if any, after the first `=` up to the end of _bytes_.

Issue #64, opened 2015-08-21
Bugs in timeout test and status code handling


maier49:
The timeout test is currently failing because it is resolving with a status code before the timeout error is thrown, adding a delay to the request that is greater than the duration of the timeout should resolve this. 

Additionally, xhr.ts is currently considering any status code less than 200 or greater than or equal to 400 as an error, it should be modified so that it only treats codes less than or equal to 0, or greater than or equal to 400 as errors.

kfranqueiro:
The reason that the timeout test fails regardless is because natively setting the XHR object's `timeout` causes it to abort when it reaches that timeout, which then fires `onreadystatechange` and hits our general error handler.  Unfortunately, XHR fires `onreadystatechange` before `ontimeout`, which makes the latter basically useless for this purpose.

The only thing I've come up with so far is to try storing a Date object when we start a request with a timeout set, and compare against that if we receive a response with `0` statusCode and empty statusText to try to determine that it's a timeout error.  I'm not sure if this is entirely foolproof though.

kfranqueiro:
I suppose another option would be to not use native `timeout` and handle aborting the request ourselves, which seems sub-optimal but at least we'd definitely know exactly when we reached a timeout, because we control it.

kfranqueiro:
Unfortunately, the Date idea isn't panning out, specifically in IE.  Chrome and FF seem to do fine with it, but IE actually ends up computing an elapsed time a few milliseconds less than the specified timeout.  I guess we might not have a choice but to move to `setTimeout` and aborting it ourselves. :/

PR #66, opened 2015-08-21
Add a delay to the timeout request and fix the range of success statuses

maier49:
This includes adding an optional delay in the echo server before resolving requests, which I had hoped would fix the timeout error. It does not, unfortunately. I can remove this if it's not useful or not needed.

### DIFF
--- a/src/request/xhr.ts
+++ b/src/request/xhr.ts
@@ -69,7 +69,7 @@ export default function xhr<T>(url: string, options: XhrRequestOptions = {}): Re
 
 				response.statusCode = request.status;
 				response.statusText = request.statusText;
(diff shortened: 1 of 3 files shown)


kfranqueiro:
It might ultimately not be needed, as I think we've discovered the crux of the issue in #64, so this probably isn't affecting anything.  Though I have a nagging feeling I should keep this open in case this becomes useful for testing when we actually fix #64.

The one thing I would comment on is that the `try`/`catch` is probably overkill.

kfranqueiro:
Turns out that nagging feeling I had was right.  I discovered that IE actually has been failing the timeout test due to the request...not timing out.  It no longer does that with this branch, and behaves consistently with the other browsers.

PR #67, opened 2015-08-21
Fix request/xhr timeout logic

### LINKED ISSUE #64: Bugs in timeout test and status code handling

kfranqueiro:
This PR is based on top of a rebased/squashed version of #66, and additionally fixes #64.

As mentioned in #64, I first tried resolving this using a timestamp comparison, but that didn't turn out too well in IE.  This works across browsers.

The only possible reservation I might have is that historically, large numbers of setTimeouts were bad news for old IE, which is why Dojo 1 had a polling interval and kept track of all in-flight XHRs instead.  This PR opts for simply setting one timeout per XHR (which is also what our code in the node provider is doing), since the code for it is a lot simpler and more straightforward.  If this proves to still be a bad thing for any modern browsers we're actually interested in supporting, we can revisit this.

### DIFF
--- a/src/request/xhr.ts
+++ b/src/request/xhr.ts
@@ -2,6 +2,8 @@ import Task from '../async/Task';
 import RequestTimeoutError from './errors/RequestTimeoutError';
 import global from '../global';
 import has from '../has';
+import { Handle } from '../interfaces';
+import { createTimer } from '../util';
 import { RequestOptions, Response, ResponsePromise } from '../request';
 import { generateRequestUrl } from './util';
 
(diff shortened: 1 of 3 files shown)


kitsonk:
:+1:

kitsonk:
Merged via 9382cce0eae6d2c48fee75f7593e46d57e9e5414

PR #71, opened 2015-09-04
Fixes for test failures

bryanforbes:
A few tests use `ArrayBuffer` which fails in `IE9`. This adds a `has` test and skips the tests if `ArrayBuffer` doesn't exist.


kitsonk:
Merged via 5c51443841b067b9ea9c95c77cc03bc4d4416835

PR #74, opened 2015-10-16
Symbol Shim

kitsonk:
This PR adds a Symbol shim which is compatible with ES6/ES2015 and offloads to the native when possible.

The shim does not augment the built-in Symbol if present, which, depending upon the browser, is missing several built-in common symbols that are part of the ES6 spec.  It might be worth considering providing those symbols.


tests/intern.ts
```diff
@@ -29,7 +29,7 @@ export const proxyUrl = 'http://localhost:9001/';
 // Note that the `build` capability will be filled in with the current commit ID from the Travis CI environment
 // automatically
 export const capabilities = {
-	'browserstack.selenium_version': '2.46.0',
+	// 'browserstack.selenium_version': '2.46.0',
```
kfranqueiro: Is this commented line really intended to be committed?

tsconfig.json
```diff
@@ -17,134 +17,137 @@
 		"./typings/tsd.d.ts"
 	],
 	"files": [
-		"./src/DateObject.ts",
```
kfranqueiro, line 20: I'm not sure why this entire array got regenerated/reordered in this commit.  Was the usual grunt task still used to generate it?

(Sorry these comments are kind of nit-picks...)

kfranqueiro:
As someone still generally uninitiated to Symbols, can you explain what this would be used for?  I had naively been under the impression that symbols were something that didn't necessarily make sense to shim.

tsconfig.json
kitsonk, line 20: The latest builds of TypeScript/atom-typescript have regenerated the file without the relatives paths... we can keep ignoring it, or update it at some point.  It wouldn't have to come in with this commit, but I just got tired of pretending the file didn't change.

tests/intern.ts
kitsonk: It did while the CI/BrowserStack is being flakey...  It is mainly a reminder that it needs to be addressed before it is landed formally.

tsconfig.json
kfranqueiro, line 20: My point is that if I run `grunt dev` inside the repository, I get a bunch of changes right back to the order this file already was in.  Shouldn't `grunt dev` typically be getting run before commits (without atom-typescript running to immediately clobber it again, I thought we had generally stopped using that)?

kitsonk:
Having a shim for Symbol enables the possibility of some other items... like shimming Iterators.  Trying to support the Iterator API becomes really silly without Symbols.  Also, TypeScript can understand symbols, but will not polyfill the functionality, so in order to write TypeScript code that uses/supports Symbols and not have significant amounts of alternative code paths, a Symbol shim effectively does that, so a single code path.  Again, Iterators are the main use case at the moment, but as more of ES6 comes to bear, many of the built ins and other objects will be created with Symbols.

On those browsers that support it (Safari 9+, Chrome, Firefox, NodeJS 4+, Opera, Edge), where Symbols are used, it reduces the overhead of processing text strings for property names (and is optimisation safe), while the non-supported browsers, it only introduces minimal overhead (and I still believe optimise safe).

tsconfig.json
kitsonk, line 20: That was after a `grunt dev`.  Whatever the `updateTsconfig` task is doing in core isn't backing out those changes.

src/Symbol.ts
```diff
@@ -0,0 +1,182 @@
...
+		return function (desc: string|number): string {
+			let postfix = 0;
+			let name: string;
+			while (created[String(desc) + (postfix || '')]) {
+				++postfix;
+			}
+			desc += String(postfix || '');
+			created[desc] = true;
+			name = '@@' + desc;
+			defineProperty(objPrototype, name, {
```
mwistrand, line 80: I can understand why it would make sense to add the so-called "well-known" symbols to `Object.prototype`, but what reason is there for decorating `Object.prototype` with _all_ shimmed symbols?

mwistrand:
With regard to providing any missing well-known symbols, I see two paths forward:
1. Decorate native `Symbol` with the missing well-knowns.
2. Rework the module so that a wrapper is always used instead of either the shim or native `Symbol` directly.

Otherwise, how useful will the other shims be without the symbols they rely on? For example, as mentioned earlier, "trying to support the Iterator API becomes really silly without Symbols." If I want to write a custom iterator that will not need to be updated once the native implementations are in place, then I will expect to be able to use the format `object[Symbol.iterator] = function () { ... };`. So I do not see how we could get away with not including support for the missing symbols as the other shims are implemented.

kitsonk:
I am going to address these comments in #115 since it seems a bit silly to update both branches at the moment.  But for the Well Known Symbols, after thinking about it, I think it is just best to fill any missing well known symbols after we have resolved if we are using the native or shim.  This in theory could create a well known symbol someone wasn't expecting, partially breaking the principal of _don't interfere with other code_ but to hold to the principal would require a lot of work for little value.

src/Symbol.ts
kitsonk, line 80: What this does is ensure that there is a setter for the symbol that makes it non-enumerable...  Otherwise you would run into this problem:

``` javascript
const obj = {};
obj[Symbol('foo')] = 'bar';
Object.keys(obj); // [ '@@foo' ]
/* instead of */
Object.keys(obj); // [ ]
```

mwistrand, line 80: That makes sense; thanks for the explanation.

kitsonk:
Closed via a20235038c53d33eccfb31bd6c69ffc5e21e882c

PR #77, opened 2015-11-17
Evented tweak

drjayvee:
I came across these two tiny issues while reading the code for Evented.

I'm not 100% about the 2nd commit (which removes the `defineProperty` call), but I cannot see any use of the property, and the tests pass fine without it.

### DIFF
--- a/src/Evented.ts
+++ b/src/Evented.ts
@@ -23,6 +23,7 @@ export default class Evented {
 	on(type: string, listener: (event: EventObject) => void): Handle {
 		const name = '__on' + type;
 		if (!(<any> this)[name]) {
+			// define a non-enumerable property (see #77)
 			Object.defineProperty(this, name, {
 				configurable: true,
 				value: undefined,
--- a/src/Evented.ts
+++ b/src/Evented.ts
@@ -29,6 +29,6 @@ export default class Evented {
 				writable: true
 			});
 		}
-		return on(this, '__on' + type, listener);
+		return on(this, name, listener);
 	}
 }


kfranqueiro:
I think your 2nd commit will end up making those `__on` methods enumerable which is a change from what it presently does.  `aspect` will still automatically add the member if it doesn't already exist, but it doesn't use `defineProperty` to do so.

We probably don't have a test for that, and one could argue it's an implementation detail, but we could probably frame a test around "this object should still only have the same enumerable members I expect even after calling `Evented#on`".

drjayvee:
Ah, I see! I'll revert that commit and add a line of documentation.

drjayvee:
There you go!

drjayvee:
The `--on<type>` method is a bit of a workaround because `on` relies on `aspect`. If only `on` [could be implemented in such a way as to not need it](https://github.com/dojo/core/issues/78)... ;)

kitsonk:
:+1: merging...

kitsonk:
One problem though is I tried to check the [CLA](http://dojofoundation.org/about/claCheck) and didn't find you under @drjayvee or the e-mail address in the commit.  Can you let me know what e-mail address/usename you used for your CLA or if you haven't you could [file one](http://dojofoundation.org/about/cla)?

kfranqueiro:
He's got a CLA.

> Jeroen Versteeg submitted a CLA on 2015-08-14 09:24:54.

drjayvee:
@kfranqueiro, can you tell me which address I used for the CLA?

I just added jeroen@asystance.nl (the address used for these commits) to my GitHub account.

kfranqueiro:
The CLA check doesn't show me e-mail addresses.  I'll see if we can find out.

kfranqueiro:
@drjayvee, FYI it looks like you used your gmail address when signing the CLA.

PR #80, opened 2015-11-20
Remove Object.observe shim

bryanforbes:
With `Object.observe()` being removed for ES2016, this should probably be removed as well.


kitsonk:
:+1: I can land if you want.

bryanforbes:
Go for it

PR #87, opened 2015-11-25
Use generics better

maier49:
In addition to adding generics to lang.ts, I updated the version of typescript and related dependencies. I'm not sure if that's something we actually want to do right now so I can roll that back if not.


kitsonk:
Sorry this took so long.  Yes, I suspect we will find more, but this is a good start.  Can you rebase?

PR #88, opened 2015-11-25
Set proper content type for xml returned from echo server

maier49:
The content type on the data being returned from the echo server was `application/octet-stream`, as we were just passing the headers from the server-side call to retrieve the xml document through to the client. IE 10 couldn't figure out that it should be treating the response as xml. Setting the content-type header in the case that the requested type is xml resolved the issue.


kitsonk:
:+1: merging now

PR #91, opened 2015-12-03
added filter to convert Buffer to JSON in node

odoe:
I'm developing parts of an API using `dojo-core/request` in Node before moving to browser development. Currently, `request` in node will return all response data as `Buffer`, even if you specify `responseType:"json"`.

I'm assuming users, even in Node, if they specify a json response, would expect json and not a buffer. So I just added another filter to account for this. I'm using `response.data.readDoubleLE` as a Buffer check. I'm not sure what other way to check than on a unique method name.

I have CLA under rrubalcava@odoe.net.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -209,7 +209,7 @@ filterRegistry.register(
  */
 filterRegistry.register(
 	function (response: Response<any>, url: string, options: RequestOptions) {
-		return response.data.readDoubleLE !== undefined && options.responseType === 'json';
+		return Buffer.isBuffer(response.data) !== undefined && options.responseType === 'json';
 	},
 	function (response: Response<any>, url: string, options: RequestOptions): Object {
 		return {
--- a/src/request.ts
+++ b/src/request.ts
@@ -208,8 +208,8 @@ filterRegistry.register(
  * Add a filter that automatically parses incoming Buffer responses in Node.
  */
 filterRegistry.register(
-	function (response: Response<any>, url: string, options: RequestOptions) {
-		return Buffer.isBuffer(response.data) !== undefined && options.responseType === 'json';
+	function (response: Response<any>, url: string, options?: RequestOptions) {
+		return options && options.responseType === 'json' && Buffer && typeof Buffer.isBuffer(response.data) !== 'undefined';
 	},
 	function (response: Response<any>, url: string, options: RequestOptions): Object {
 		return {
--- a/src/request.ts
+++ b/src/request.ts
@@ -203,3 +203,17 @@ filterRegistry.register(
 		};
 	}
 );
+
+/**
+ * Add a filter that automatically parses incoming Buffer responses in Node.
+ */
+filterRegistry.register(
+	function (response: Response<any>, url: string, options: RequestOptions) {
+		return response.data.readDoubleLE !== undefined && options.responseType === 'json';
+	},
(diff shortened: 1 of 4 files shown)


dylans:
Thanks for the PR @odoe , we'll review this asap!

kitsonk:
> Rene Rubalcava (odoe) submitted a CLA on 2015-04-16 15:33:45.

kitsonk:
@odoe thanks a lot for the PR.  A couple of thoughts:
- Would a better guard/check be `request.data instanceOf Buffer`?
- Would it be possible to get a unit test case for this?

odoe:
Thanks @kitsonk 

I updated the check to use `Buffer.isBuffer(response.data)` and added a unit test for the buffer filter.

tests/unit/request.ts
```diff
@@ -224,6 +224,22 @@ if (has('host-node')) {
...
 				);
+		},
+
+		'Buffer filter'() {
+			handle = filterRegistry.register(/foo\.json$/, function (response: Response<any>) {
+				return response;
+			});
+
+			const dfd = this.async();
+      const options: RequestOptions = { responseType: 'json' };
```
kitsonk: Some sort of indentation issue, but I can fix this when I merge it.

src/request.ts
```diff
@@ -203,3 +203,17 @@ filterRegistry.register(
 		};
 	}
 );
+
+/**
+ * Add a filter that automatically parses incoming Buffer responses in Node.
+ */
+filterRegistry.register(
+	function (response: Response<any>, url: string, options: RequestOptions) {
+		return Buffer.isBuffer(response.data) !== undefined && options.responseType === 'json';
```
kitsonk: I am afraid you are accessing Buffer in an unsafe way, which is [throwing errors](https://travis-ci.org/dojo/core/builds/95615947#L388) in browser environments.  Also, we should consider narrowest to widest in our logic here to take advantage of short cutting.

I think a better way would be something like this:

``` typescript
return options.responseType === 'json' && Buffer && typeof Buffer.isBuffer(response.data) !== 'undefined';
```

kitsonk: Also, I just looked at the code, the technically, `options` is an optional argument, so we should change the `options: RequestOptions` to `options?: RequestOptions` and safely access `options` as well.

odoe:
Updates made @kitsonk. I'm getting a failing test in `request/xhr.js` in travis, but not sure if it's related to my updates or not. From what I can see, I don't think it is.

Do you think it's necessary to be even more explicit and wrap the register method for buffer in a `has('host-node')`?

Thanks!

kitsonk:
I double and triple checked, and it isn't related to this PR... it is currently the same behaviour in master...  It is really strange and I can't seem to pin it down, because it only occurs when we use the internal reporter, but not the standard Intern ones.  I will chase it down.

If we wanted to improve it, instead of feature detecting has('host-node'); I would create a `core/request/has` and do something like this:

``` typescript
import has, { add } from '../has';

add('node-buffer', (global) => { 'Buffer' in global && typeof global.Buffer === 'function'; });

export default has;
```

And then in the module `import has from './request/has';` and then wrap the registration in a `has('node-buffer');` and you can then probably simplify the logic in the matcher then too...

odoe:
Cool, added `core/request/has` and updated `core/request`. Tests look good on this one.

Thanks!

PR #96, opened 2015-12-11
Account for slow environments in tests that utilize `setTimeout`.

### LINKED ISSUE #85: Promise fulfilled delay failures
We are routinely getting something like the following.  We need to figure out if the test is just being temperamental or if there is some problem we are regularly overrunning this timeout:

```
× firefox on Windows 10 - unit tests - async/timing - delay() - resolves a promise after the given timeout (0.465s)
AssertionError: expected 462 to be close to 250 +/- 25
  at AssertionError  <__intern/node_modules/chai/chai.js:4760:18>
  at [2]</module.exports/Assertion.prototype.assert  <__intern/node_modules/chai/chai.js:203:1>
  at [4]</module.exports/<  <__intern/node_modules/chai/chai.js:1727:1>
  at [9]</module.exports/ctx[name]  <__intern/node_modules/chai/chai.js:3475:18>
  at [5]</module.exports/assert.closeTo  <__intern/node_modules/chai/chai.js:2968:5>
  at ["delay()"]["resolves a promise after the given timeout"]/<  <_build/tests/unit/async/timing.js:21:21>
```

mwistrand:
Resolves #85. Replaces usage of `assert.closeTo` with `assert.isAbove` in tests that use `setTimeout` to measure delays. Initial attempts involved finding a more accurate delta, but as there is no way to pin down an accurate delta across all enviroments, using `isAbove` seems to be the "less accurate, but more correct" test method.

Finally, since Intern uses Chai v3, the corresponding d.ts file in this package has been updated with these typings from the older 2.0.0 typings.

### DIFF
--- a/tests/typings/chai/chai.d.ts
+++ b/tests/typings/chai/chai.d.ts
@@ -1,9 +1,12 @@
-// Type definitions for chai 2.0.0
+// Type definitions for chai 3.2.0
 // Project: http://chaijs.com/
 // Definitions by: Jed Mao <https://github.com/jedmao/>,
 //                 Bart van der Schoor <https://github.com/Bartvds>,
-//                 Andrew Brown <https://github.com/AGBrown>
+//                 Andrew Brown <https://github.com/AGBrown>,
+//                 Olivier Chevet <https://github.com/olivr70>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
+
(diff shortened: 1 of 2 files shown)


mwistrand:
Intern uses Chai v3.0.0, but the typings provided by DefinitelyTyped are for v3.2.0. The differences are outlined [here](https://github.com/dojo/dojo2-package-template/issues/21).

PR #97, opened 2015-12-11
Load local modules before running test.

### LINKED ISSUE #82: Contextual Load Failures
We are repeatedly getting the following test failures:

```
CancelError: Timeout reached on internet explorer 11.0 on Windows 7 - unit tests - load - contextual load
CancelError: Timeout reached on internet explorer 10.0 on Windows 7 - unit tests - load - contextual load
CancelError: Timeout reached on chrome on Windows 10 - unit tests - load - contextual load
```

maier49:
Fixes #82

### DIFF
--- a/tests/unit/load.ts
+++ b/tests/unit/load.ts
@@ -7,6 +7,10 @@ import Promise from 'src/Promise';
 const suite: any = {
 	name: 'load',
 
+	before() {
+		return load('tests/unit/load/a', 'tests/unit/load/b')
+	},
+
 	'global load'() {
 		const def = this.async(5000);
 


maier49:
These failures seem to be caused by require just taking a long time to load these modules. I was able to reproduce them intermittently locally and consistently on BrowserStack and made the following observations:
- This error only happens when running a lot of tests(i.e. just running the load tests or the load tests and one or two other modules usually doesn't produce the error, either locally or on BrowserStack)
- Locally, removing the echo server and hitting the intern proxy directly reduces the time for this test to on average < 100ms and I never see this error, but even with this change the error still happens on BrowserStack.

Unfortunately that's not very conclusive at all.
This fix avoids the errors by pre-loading the modules, and after testing on BrowserStack and locally does not fail in IE 10 or 11, Firefox, or Chrome. So it's not really a fix so much as a workaround. However, the 'load' tests do take quite a while to run since they first wait for the modules to be loaded.

In the meantime I'll keep trying to identify the underlying cause.

Edit: The tests still fail due to a timeout after replacing the contents of the tests with calls to load the same files using xhr. So it may represent a problem with the performance of the tests but doesn't appear to be a problem with the load module itself.

PR #98, opened 2015-12-11
Add guard for aborted requests.

### LINKED ISSUE #94: IE9 posting coverage...
For some reason we seem to be [having issues with IE9](https://travis-ci.org/dojo/core/builds/95359212#L199) at the end of its test in CI.  A quick google appears to be related to throwing an error when trying to introspect XMLHttpRequest at the wrong time.  I am not sure if it is something in dojo/core or something in Intern that is causing the challenge.

mwistrand:
Resolves #94. Adds a guard to the `onreadystatechange` callback so that it only proceeds when the request has not been aborted. As a result, aborted requests will not throw errors in IE9.

### DIFF
--- a/src/request/xhr.ts
+++ b/src/request/xhr.ts
@@ -40,6 +40,12 @@ export default function xhr<T>(url: string, options: XhrRequestOptions = {}): Re
 			return request.getResponseHeader(name);
 		}
 	};
+	let isAborted: boolean = false;
+
+	function abort() {
+		isAborted = true;
+		request && request.abort();
+	}
 
 	const promise = new Task<Response<T>>(function (resolve, reject): void {
 		if (!options.method) {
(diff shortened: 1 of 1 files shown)


kitsonk:
Should there be a specific unit test for this, or do you think the existing tests already cover this off fully?

kitsonk:
Nevermind... I see you comment in #94 about the unit tests.

src/request/xhr.ts
```diff
@@ -60,7 +66,7 @@ export default function xhr<T>(url: string, options: XhrRequestOptions = {}): Re
 
 		let timeoutHandle: Handle;
 		request.onreadystatechange = function (): void {
-			if (request.readyState === 4) {
+			if (!isAborted && request.readyState === 4) {
```
kitsonk, line 69: Shouldn't we handle (as in reject) `isAborted && request.readyState === 4)`?  Because obviously that is happening at runtime and that should result in a rejected promise, right?  With something like `resolve(new Error('An aborted request has been resolved'));`

mwistrand, line 69: If the request times out, then the underlying promise is rejected. However, since the `request` object itself is unaware of any promise state, it fires that `onreadystatechange` callback anyway. But you're right; when `Task#cancel` is executed, the task itself is not rejected. So I wonder if rejecting the promise as aborted within the local `abort` function and resetting `onreadystatechange` to a noop function is the way to go here.

kitsonk, line 69: While the `request` object is unaware of the promise state, it is within closure that has the related promise, named as `promise`, but I agree that the best thing to do is reject and set to a noop.

PR #108, opened 2016-01-17
Add handleAs to JSON filter on request.

kitsonk:
This refactors the filter logic to accept `options.responseType` or `options.handleAs` as values for the filter (and rewrites the test to actually test the filter logic, which they were illogically not testing actually).

Also, simply coerced `response.data` to `String()` in filter to deal with node buffers.  This eliminates the need for a separate filter for NodeJS as previously.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -199,7 +199,7 @@ filterRegistry.register(
 	},
 	function (response: Response<any>, url: string, options: RequestOptions): Object {
 		return {
-			data: JSON.parse(response.data)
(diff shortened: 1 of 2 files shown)


odoe:
It doesn't look like the string coercion made it into this commit.

kitsonk:
You are right.  Fixed now... Also changed the tests so mock emit buffers under node, which should prevent that omission from occurring again.  The tests in this area still need some improvement.  They are a bit hacky.

PR #113, opened 2016-02-27
add array#includes method and tests

nicknisi:
This PR adds an Array includes method implementation and tests.

### DIFF
--- a/src/array.ts
+++ b/src/array.ts
@@ -208,3 +208,23 @@ export function copyWithin<T>(target: ArrayLike<T>, offset: number, start?: numb
 
 	return target;
 }
+
+/**
+ * Determines whether an array includes a given value
+ * @param target the target array-like object
+ * @param searchElement the item to search for
+ * @param fromIndex the starting index to search from
+ */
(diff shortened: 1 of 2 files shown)


kitsonk:
Why no offloading to native?

kfranqueiro:
@kitsonk you could ask the same of all of the other functions in `array.ts`.

I suspect that originally the answer to this question was because we originally proposed also having the option of hooking these methods up as polyfills, but we eventually [put the brakes on that idea](https://docs.google.com/document/d/1E_nh2Cq0E6LJCXPvL2mbPujrpyBIXSXTr2epwlPVABU/edit#heading=h.hx1cbuyjakf4).

Either way, this PR is currently consistent with everything else in `array.ts` - the question of whether to defer to native when present is probably a larger separate question.

nicknisi:
Yeah, I kept it consistent with the other methods in this file, but would be happy to add native offloading to these methods.

kitsonk:
Well since this matches existing, lets just merge it and open up another issue to do native offloading (#114)

kitsonk:
:+1: from me

nicknisi:
Merged via e97df09

PR #115, opened 2016-02-29
Shim module for Set

kitsonk:
This module requires #74 and shims the functionality of `Set`, which offloads to native (when there is a compliant es6 `Set` implementation, IE 11 and older versions of Safari do not provide sufficient functionality, and so the shim is used in those situations).


kitsonk:
I should also mention that it includes a module named `iterator` which provides a shim for common iterator type functionality, including the ability to replicate the `for ... of` type of iteration via a `forOf()` function.  Where you can take something that defines a [Symbol.iterator] or iterators over anything that is `ArrayLike`, for example:

``` typescript
import { forOf } from 'dojo-core/iterator`;

const arr: number[] = [ 1, 2, 3 ];
forOf(arr, function (value, source, doBreak) {
  console.log(value);
  /* source is the original iterator object */
  /* doBreak() is a function which allows you to break the forOf loop */
});
```

Since it isn't possible to "shim" syntax, it isn't possible to offload to native fully, but it does grab whatever iterator might be available and iterates over that until done.

kitsonk:
So, it appears that TypeScript treats `Symbol` as a special object and elides it from typing information in a declaration.  So even with an explicit import it still got elided.  It would then cause issues in that the resulting `.d.ts` in the distribution would have issues with `Symbol` not being found.  The only work around I was able to get to work was to provide a global Symbol `.d.ts` file which consumers will need to include in the scope of the project (or include `lib.es6.d.ts` or any other polyfil that declares `Symbol` in the global scope).  At that point, the typings will be correct and the run-time behaviour will be preserved.

src/Set.ts
```diff
@@ -0,0 +1,118 @@
+import global from './global';
+import Symbol from './Symbol';
+import { IterableIterator, Iterable, ShimIterator } from './iterator';
+import { hasClass } from './decorators';
+import { forOf } from './iterator';
```
mwistrand: This currently imports from `./iterator` twice.

src/Set.ts
```diff
@@ -0,0 +1,118 @@
...
+			if (idx === -1) {
+				return false;
+			}
+			this._setData.splice(idx, 1);
+			return true;
+		};
+
+		entries(): IterableIterator<[T, T]> {
+			const ent = new ShimIterator(this._setData.map((value) => [ value, value ]));
+			(<any> ent)[Symbol.iterator] = function (): IterableIterator<[T, T]> {
```
mwistrand: As far as I can tell, none of Dojo 2's target environments support just `Symbol` without `Symbol.iterator`, so this might work in all target environments without a hitch.

src/Set.ts
```diff
@@ -0,0 +1,118 @@
...
+		[Symbol.iterator](): IterableIterator<T> {
+			const iterator = new ShimIterator(this._setData);
+			(<any> iterator)[Symbol.iterator] = function (): IterableIterator<T> {
+				return <any> iterator;
+			};
+			return <any> iterator;
+		};
+
+		[Symbol.toStringTag]: string = 'Set';
+	}
```
mwistrand, line 71: Since `Set.prototype[@@iterator]` is the [same function as Set.prototype.values](http://www.ecma-international.org/ecma-262/6.0/#sec-set.prototype-@@iterator), and since the implementation is currently identical (apart from the `iterator` variable name), this duplicated method could be shortened.

src/iterator.ts
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	/* We need to handle iteration of double byte strings properly */
+	if (!isIterable(iterable) && typeof iterable === 'string') {
+		const l = iterable.length;
+		for (let i = 0; i < l; ++i) {
+			let char = iterable[i];
+			if ((i + 1) < l) {
+				const code = char.charCodeAt(0);
+				if ((code >= 0xD800) && (code <= 0xDBFF)) {
```
mwistrand: These characters are currently exported by `dojo/string` as the constants `HIGH_SURROGATE_MIN` and `HIGH_SURROGATE_MAX`, respectively.

src/Set.ts
mwistrand, line 71: Also, `[Symbol.toStringTag]` will break in environments that support `Symbol` but not `Symbol.toStringTag` (e.g., the current version of Firefox (45.0.1)), as the `dojo/Symbol` currently does not add missing well-knowns.

kitsonk, line 71: Good point, the shim for Symbol should do all the ES6 well knowns.

Also, will look at your other comment.

src/iterator.ts
kitsonk: Awesome, thanks!

src/Set.ts
kitsonk: Ok, I think I figured out what your point was.  When I do a next push, it will hopefully address this point.

kitsonk:
@mwistrand if you could take one more look, specifically at the changes I made (and any outstanding feedback you don't think I addressed) and then I think I can land this.  I didn't directly update the Symbol branch (#74) as I just figured I would do it all in this PR.

mwistrand:
Thanks, @kitsonk, for your responses and changes. Any concern/suggestion I had has been addressed, for both this and #74.

kitsonk:
Closed via a20235038c53d33eccfb31bd6c69ffc5e21e882c

PR #118, opened 2016-03-08
Update to use latest dojo-loader with new interfaces

tomdye:
Requires: https://github.com/dojo/loader/pull/53

Fixed to use the latest typings form the new `dojo-loader`. This requires a release of `dojo-loader` with the globally declared `define` and `require` interfaces.
- Fixes issues in `on/nodeOnly` and `request` unit tests
- Moves node and browser specific `request` tests into their own files

PR #120, opened 2016-03-09
Text plugin - Fixes tests and updates code for latest loader

tomdye:
Fixes: https://github.com/dojo/loader/issues/4
Dependencies: https://github.com/dojo/core/pull/118 and latest `dojo-loader`
- Port of dojo1.x text loader plugin by @bitpshr 
- Adds config to load function (optional)
- Adds unit tests


tomdye:
@kitsonk updated from https://github.com/dojo/core/pull/118. Can be merged afterwards. This should enable `intern` to use latest loader / core

Issue #27, opened 2015-05-28
Implement has loader plugin


kfranqueiro:
`dojo-core/has` has been implemented, but the loader plugin APIs have not been implemented to allow it to be used for conditionally loading module dependencies.

Conditionally-loaded dependencies look like the following:
- `dojo-core/has!feature?featureModule:noFeatureModule` to load `featureModule` if `feature` is truthy, otherwise load `noFeatureModule`
- `dojo-core/has!feature?featureModule` to load `featureModule` if `feature` is truthy, otherwise load nothing
- `dojo-core/has!feature?:noFeatureModule` to load `noFeatureModule` if `feature` is falsy, otherwise load nothing

We aren't sure yet how much the loader plugin API may deviate from the current AMD plugin API, so implementing this needs to wait until that is settled.  (Or it could initially be developed as a test for whatever new API is taking shape.)

bryanforbes:
As per dojo/loader#9, the loader will stick with the AMD plugin API. The `has` module needs to be updated to add this.

kfranqueiro:
It has occurred to me (after seeing additional comments on dojo/loader#9) that dynamic module loading in general is likely to be a royal pain in the rear-end, whether `has` is implemented as a plugin or not.  Maybe not as unfathomable as the dynamic loading going on in `request` right now, for example, but still not exactly painless to write.  I'm/we're going to have to think about how this is going to end up presenting to end-users.

kitsonk:
Yes, but it is a very powerful feature and it needs to be thought about in the full end-to-end tool chain too.  While expressing conditional loading at the module level is necessary, it should seldom be conditional at final run-time as to deliver efficient code, it should be built out at concatenation/build.

(I think) what we like about AMD, which is less the "asynchronous" but the more robust modularisation than was available in CommonJS and what has been delivered in ES6, but modularisation that allowed emitting concise optimised layers.  AMD, has and the build tool in Dojo 1 was pretty "feature complete" though, ironcially, the ability for it all to work functionally the same without the build step let a lot of end-developers to never build optimised and led a level of criticism that Dojo 1 was "slow".

kfranqueiro:
Yeah, that's a good point, either way we will also need to accommodate resolving this stuff at build time like Dojo 1 does.

@bryanforbes suggested we should be able to hide _most_ of the dynamic-loading complexity behind an API with something like this (essentially reduced from the stuff in `request`):

``` js
function getDeps(deps) {
[...]

PR #122, opened 2016-03-14
#27:  Add has plugin

tomdye:
Fixes: https://github.com/dojo/core/issues/27
Brings in `has` code from @jdonaghue branch.


src/has.ts
```diff
@@ -76,20 +132,24 @@ add('weakmap', function () {
 		const map = new global.WeakMap([ [ key1, 1 ] ]);
 		return map.get(key1) === 1 && map.set(key2, 2) === map;
 	}
+	/* istanbul ignore next: node environment will never get here */
```
kitsonk: Why are we doing this?  When the coverage is run in a combined CI way, we will cover this.

tomdye: I didn't notice that line, was pulled in from the previous branch

kitsonk: There are quite a few of them... I would drop them, since our testing now properly covers these sorts of things.

PR #130, opened 2016-03-18
Use native array objects when available

### LINKED ISSUE #114: Offloading to native for array.ts
We should offload to native functionality for array when the native functions are available.

devpaul:
array now uses native objects by default when available. Fixed a small bug for non-integer end values in `copyWithin()` and `fill()`.

 Fixes #114

### DIFF
--- a/src/array.ts
+++ b/src/array.ts
@@ -70,7 +70,7 @@ export function from<T>(arrayLike: ArrayLike<T>, mapFunction?: MapCallback<T>, t
  */
 export function from<T>(arrayLike: (string | ArrayLike<T>), mapFunction?: MapCallback<T>, thisArg?: {}): ArrayLike<T> {
 	// Use the native Array.from() if it exists
-	if (has('array-from')) {
+	if (has('es6-array-from')) {
 		return (<any> Array).from.apply(null, arguments);
 	}
 
@@ -106,7 +106,7 @@ export function of(...items: any[]): any[];
  * @return An array from the given arguments
  */
 export function of() {
-	if (has('array-of')) {
+	if (has('es6-array-of')) {
 		return (<any> Array).of.apply(null, arguments);
 	}
 
@@ -123,7 +123,7 @@ export function of() {
  * @return The filled target
  */
 export function fill<T>(target: ArrayLike<T>, value: any, start?: number, end?: number): ArrayLike<T> {
-	if (has('array-fill')) {
+	if (has('es6-array-fill')) {
 		const method = (<any> Array.prototype).fill;
 		return method.call.apply(method, arguments);
 	}
@@ -149,7 +149,7 @@ export function fill<T>(target: ArrayLike<T>, value: any, start?: number, end?: 
  * @return The first index whose value satisfies the passed callback, or -1 if no values satisfy it
  */
 export function findIndex<T>(target: ArrayLike<T>, callback: FindCallback<T>, thisArg?: {}): number {
-	if (has('array-findIndex')) {
+	if (has('es6-array-findIndex')) {
 		const method = (<any> Array.prototype).findIndex;
 		return method.call.apply(method, arguments);
 	}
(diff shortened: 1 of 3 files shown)


src/array.ts
```diff
@@ -113,9 +123,14 @@ export function of() {
...
 export function fill<T>(target: ArrayLike<T>, value: any, start?: number, end?: number): ArrayLike<T> {
+	if (has('array-fill')) {
+		const method = (<any> Array.prototype).fill;
+		return method.call.apply(method, arguments);
+	}
+
 	const length = toLength(target.length);
 	let i = normalizeOffset(toInteger(start), length);
-	end = normalizeOffset(end ? toInteger(end) : length, length);
+	end = normalizeOffset(end === undefined ? length : toInteger(end), length);
```
devpaul, line 133: The spec only defaults to length when end is undefined and uses toInteger() for all other values: http://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.fill. Caught this issue thanks to leveraging the dojo array tests for the native array passthrough.

src/array.ts
```diff
@@ -183,7 +213,7 @@ export function copyWithin<T>(target: ArrayLike<T>, offset: number, start?: numb
 	const length = toLength(target.length);
 	offset = normalizeOffset(toInteger(offset), length);
 	start = normalizeOffset(toInteger(start), length);
-	end = normalizeOffset(end ? toInteger(end) : length, length);
+	end = normalizeOffset(end === undefined ? length : toInteger(end), length);
```
devpaul, line 216: Same rule here as well

src/array.ts
```diff
@@ -68,6 +69,11 @@ export function from<T>(arrayLike: ArrayLike<T>, mapFunction?: MapCallback<T>, t
  * @return The new Array
  */
 export function from<T>(arrayLike: (string | ArrayLike<T>), mapFunction?: MapCallback<T>, thisArg?: {}): ArrayLike<T> {
+	// Use the native Array.from() if it exists
+	if (has('array-from')) {
```
kitsonk: While we don't have it standardized yet (see dojo/meta#5) I would still like to try to use them so could we do `es6-array-from` _et al._ instead?

src/array.ts
kitsonk, line 133: Good catch!

kitsonk:
Also to note, we haven't figured out how to build optimize `has()` features yet and if whatever from of tree-shaking we will do will elide properly.  So we may have to revisit.

Is it worth considering the logic order too?  Presuming native and only providing functionality when native isn't there?

src/has.ts
```diff
@@ -169,10 +169,10 @@ add('xhr2-blob', function () {
...
-add('array-find', 'find' in global.Array.prototype);
-add('array-copy-within', 'copyWithin' in global.Array.prototype);
-add('array-includes', 'includes' in global.Array.prototype);
+add('es6-array-from', 'from' in global.Array);
+add('es6-array-of', 'of' in global.Array);
+add('es6-array-fill', 'fill' in global.Array.prototype);
+add('es6-array-findIndex', 'findIndex' in global.Array.prototype);
+add('es6-array-find', 'find' in global.Array.prototype);
+add('es6-array-copyWithin', 'copyWithin' in global.Array.prototype);
+add('es7-array-includes', 'includes' in global.Array.prototype);
```
devpaul, line 178: marked `includes` as es7 rather than esnext. Don't let me down TC39!

src/array.ts
```diff
@@ -200,8 +200,8 @@ export function find<T>(target: ArrayLike<T>, callback: FindCallback<T>, thisArg
  * @param end The last (exclusive) index to copy; if negative, it counts backwards from length
  * @return The target
  */
-export function copyWithin<T>(target: ArrayLike<T>, offset: number, start?: number, end?: number): ArrayLike<T> {
-	if (has('array-copyWithin')) {
+export function copyWithin<T>(target: ArrayLike<T>, offset: number, start: number, end?: number): ArrayLike<T> {
```
devpaul, line 203: According to the spec, start is required, even though undefined will evaluate to 0. ¯_(ツ)_/¯

src/has.ts
kitsonk, line 178: It is officially in ES2016/7, only that and the exponentiation operator. :grinning:

PR #141, opened 2016-04-06
meta #25: Compile ES6 version of core

tomdye:
### Issue

Addresses: https://github.com/dojo/meta/issues/25
### Pre-reqs

Requires Intern release and `nodeOptions` pr merge: https://github.com/theintern/intern/pull/604#
### New grunt tasks
- `grunt dev-es6` - runs `dev` task with `tsOptions.target` set to `es6`
- `grunt test-es6` - runs `test` task with `tsOptions.target` set to `es6` and node `harmony` flags enabled
- `grunt dist-es6` - runs `dev` task with `tsOptions.target` set to `es6`


tests/unit/Promise.ts
```diff
@@ -310,7 +310,7 @@ export function addPromiseTests(suite: any, Promise: PromiseType) {
 				let dfd = this.async();
 				let error = new Error('foo');
 				let foreign = <any> {
-					then: function (f: Function) {
+					then: (f: Function) => {
```
Tomdye: Attempting to fix an error, didn't work.
Test not returning `Error: Foo`, instead `ReferenceError: this is not defined`

tests/unit/array.ts
```diff
@@ -10,14 +10,17 @@ function assertFrom(arrayable: any, expected: any[]) {
...
 
-class MyArray {
-	constructor() {
-		Array.apply(this, arguments);
-		return this;
-	}
-	static from = array.from;
-}
-MyArray.prototype = Object.create(Array.prototype);
+// This breaks in the ES6 version for chrome,
```
tomdye: This chunk and it's associated test attempts to overwrite read-only prototype, breaks Chrome

tests/unit/Promise.ts
bryanforbes: What about:

``` ts
let foreign = <any> {
    then(f: Function) {
        throw error;
    }
};
```

matt-gadd:
@Tomdye what are we doing about these PR's? are they still relevant?

tomdye:
Pretty stale now to be honest, the `systemjs` compat stuff probably isn't needed anymore. Might be worth closing and re-raising the `es6` fixes, though they're probably already done elsewhere.

agubler:
yeah really stale now given it has it's own grunt file and doesn't use grunt-dojo2

Issue #137, opened 2016-03-29
Check if this applies to new dojo-core/request: node request handler shouldn't normalize headers


tomdye:
@Tomdye commented on [Mon Mar 07 2016](https://github.com/dojo/loader/issues/52)
Related to `request/xhr` and `request/node`

dojo2 issue: https://github.com/dojo/dojo2/issues/17

PR #150, opened 2016-04-21
#137: Headers should not be Normalised

tomdye:
Fixes: https://github.com/dojo/core/issues/137

No longer normalises headers in `request/node.ts`


kitsonk:
I know it sounds a bit silly, but if normalizing headers breaks things, even though we removed functionality, could we have a test to validate that headers do not get normalized?  That way we don't accidentally regress and are explicit about the fact that is now the functionality.

tomdye:
Sure thing, guess that makes sense, i'll re-instate the former test and adjust the assertions

PR #151, opened 2016-04-21
Update to allow testing against Sauce Labs and Browserstack

michael-vansickle:
Updated Gruntfile and intern configurations to allow tests to run against saucelabs and browserstack via grunt command line settings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ install:
 script:
   - "grunt"
   - "grunt intern:node --combined"
-  - "grunt intern:remote --combined"
+  - "grunt intern:saucelabs --combined"
   - "grunt remapIstanbul:ci"
   - "grunt uploadCoverage"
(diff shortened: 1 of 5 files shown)


kitsonk:
Can you switch over to SauceLabs before I merge this one?

typings/browser.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="browser/ambient/jsdom/index.d.ts" />
```
kitsonk: Why is this here, it seems tangental to the issue being addressed.

michael-vansickle:
I've updated the .travis.yml file to use Saucelabs. I also cleaned out the old jsdom stuff that made its way in when I was experimenting with something.

kitsonk:
Merged in ee8949753f562abab17fed29717609c16703a947

PR #152, opened 2016-04-22
Add isTask type guard to async/Task.

kitsonk:
Adds a Task type guard to async/Task

### DIFF
--- a/tests/unit/async/Task.ts
+++ b/tests/unit/async/Task.ts
@@ -9,10 +9,9 @@ let suite = {
 
 	'isTask()'() {
 		const task = new Task((resolve) => resolve(), () => {});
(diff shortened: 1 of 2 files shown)


tests/unit/async/Task.ts
```diff
@@ -1,12 +1,25 @@
...
+import Task, { Canceled, isTask } from 'src/async/Task';
 import Promise from 'src/Promise';
 import { addPromiseTests } from '../Promise';
 
 let suite = {
 	name: 'Task',
 
+	'isTask()'() {
+		const task = new Task((resolve) => resolve(), () => {});
+		const promise = new Promise((resolve) => resolve());
```
novemberborn: Nit: `Promise.resolve()` should suffice.

PR #154, opened 2016-04-25
Task.resolve() returns a Task

### LINKED ISSUE #153: async/Task.resolve() does not return Task
`Task.resolve()` does not return a `Task()` [as contracted](https://github.com/dojo/core/blob/a7565bb2fd5add72c92810c6b7931e78139afd92/src/async/Task.ts#L21).  Instead it returns a `Promise`.

Because it is simply calling the `super` static method, which returns a `Promise`.  It probably should do what the `PromiseShim` [does](https://github.com/dojo/core/blob/a7565bb2fd5add72c92810c6b7931e78139afd92/src/Promise.ts#L122) instead, which is return a `new this` which is resolved with the value passed.

kitsonk:
Fixes #153

### DIFF
--- a/tests/unit/async/Task.ts
+++ b/tests/unit/async/Task.ts
@@ -147,6 +147,19 @@ let suite = {
 
 			resolver();
 		}
+	},
+
+	'Task#resolve': {
+		'returns a task'() {
+			const task = Task.resolve('foo');
+
+			assert.isFunction(task.cancel, 'A task should have a cancel function');
+			assert.isFunction(task.finally, 'A task should have a finally function');
+
+			return task.then((result) => {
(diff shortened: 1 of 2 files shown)


tests/unit/async/Task.ts
```diff
@@ -147,6 +147,19 @@ let suite = {
 
 			resolver();
 		}
+	},
+
+	'Task#resolve': {
```
novemberborn: `Task.resolve`?

kitsonk: I was following the other patterns in the file, but I guess it is a static method, versus the `#` I guess indicating an instance method.

PR #155, opened 2016-04-29
Incorporate iterators throughout core.

mwistrand:
- Allow an iterable to be passed to `array.from`.
- Convert `Map` to a shim that allows iterables and offloads to native.
- Allow iterables to be passed to the WeakMap constructor, and add
  `Symbol.toStringTag` definition.
- Add iterators to Promise and Task implementations.
- Add iterators to async/iteration.

### DIFF
--- a/src/Symbol.ts
+++ b/src/Symbol.ts
@@ -61,11 +61,17 @@ export namespace Shim {
 			desc += String(postfix || '');
 			created[desc] = true;
 			name = '@@' + desc;
-			defineProperty(objPrototype, name, {
-				set: function (value: any) {
-					defineProperty(this, name, getValueDescriptor(value));
-				}
-			});
+
+			// FIXME: Temporary guard until the duplicate execution when testing can be
+			// pinned down.
+			if (!Object.getOwnPropertyDescriptor(objPrototype, name)) {
+				defineProperty(objPrototype, name, {
+					set: function (value: any) {
+						defineProperty(this, name, getValueDescriptor(value));
+					}
+				});
+			}
+
 			return name;
 		};
 	}());
--- a/src/array.ts
+++ b/src/array.ts
@@ -1,5 +1,6 @@
-import { MAX_SAFE_INTEGER as maxSafeInteger } from './number';
 import has from './has';
+import { forOf, isArrayLike, isIterable, Iterable } from './Iterator';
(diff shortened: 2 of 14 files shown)


src/Symbol.ts
```diff
@@ -61,11 +61,17 @@ export namespace Shim {
...
 			name = '@@' + desc;
-			defineProperty(objPrototype, name, {
-				set: function (value: any) {
-					defineProperty(this, name, getValueDescriptor(value));
-				}
-			});
+
+			// FIXME: Temporary guard until the duplicate execution when testing can be
+			// pinned down.
+			if (!Object.getOwnPropertyDescriptor(objPrototype, name)) {
```
mwistrand, line 67: When running `$ grunt test`, the loader executes certain factories twice, resulting in the well-known symbols being set twice. Once I track this down, I will remove this guard.

src/array.ts
```diff
@@ -1,5 +1,6 @@
-import { MAX_SAFE_INTEGER as maxSafeInteger } from './number';
 import has from './has';
+import { forOf, isArrayLike, isIterable, Iterable } from './Iterator';
+import { MAX_SAFE_INTEGER as maxSafeInteger } from './number';
 
 export interface ArrayLike<T> {
```
mwistrand, line 5: Is `interface ArrayLike<T>` needed, since it's provided by TypeScript in lib.core.d.ts?

src/array.ts
```diff
@@ -83,17 +84,23 @@ export function from<T>(arrayLike: (string | ArrayLike<T>), mapFunction?: MapCal
 	}
 
 	const Constructor: any = this;
```
mwistrand, line 86: It's not entirely clear to me why `Constructor` is set to `this`, since `this` does not actually represent a sub-classed array. Does this assume that `from` is executed as `from.call(SubclassedArray, arrayLike)`?

src/Symbol.ts
kitsonk, line 67: Hmmm.  Strange... ok...

src/array.ts
kitsonk, line 86: I will admit, I am not sure either, it could easily have been some code "rot".  I think we can merge things and open another ticket to work on trying to clean this up a bit.

src/array.ts
kitsonk, line 5: Opened #158 to deal with this.

PR #165, opened 2016-05-23
Realign Distributions

kitsonk:
This PR realigns the package format to make it easier to build from source, but also migrate to typings/typings for managing of the typings.


tomdye:
Nice

PR #173, opened 2016-06-22
Enhance has() API

### LINKED ISSUE #172: Unregistered has() features fail silently...
Three challenges with the `has()` API:
- If a feature is not registered, it just fails silently returning `undefined`.  This leads to potential logic errors, assuming a feature is being properly detected when in fact it isn't.  We should throw.
- Features are case sensitive, and while by convention, we keep labels all lower case, consumers of the API may unintentionally make mistakes... if we coerced the features to lowercase, it would avoid mistakes and errors.
- When `add` is passed with `override = false` it simply returns a `false` if the `add` fails because it is already present.  The way we currently consume the API, we don't handle those failures and it is likely that those downstream using the API will do the same.  We should consider throwing when the override fails, as it is clearly again something someone should be explicit about, instead of unintentional behaviour.

matt-gadd:
fixes #172

### DIFF
--- a/tests/unit/has.ts
+++ b/tests/unit/has.ts
@@ -46,15 +46,29 @@ registerSuite({
 			}
 		},
 
+		'throws when no value in cache and no test'() {
+			const feature = 'abc';
+			let err: Error;
+			try {
+				has(feature);
+			} catch (e) {
+				err = e;
+			}
+			assert.isTrue(err instanceof Error);
+			assert.equal(err.message, 'abc does not exist');
+		},
+
+		'feature name is lowercased'() {
+			hasAdd('abc', true);
+			assert.isTrue(has('ABC'));
+		},
+
 		'has.cache': {
 			'basic true/false tests'() {
 				hasAdd('abc', true);
 				assert.isTrue(hasCache['abc']);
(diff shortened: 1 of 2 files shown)


tests/unit/has.ts
```diff
@@ -112,9 +126,21 @@ registerSuite({
...
+				try {
+					assert.isFalse(hasAdd('abc', false));
+				} catch (e) {
+					err = e;
+				}
+				assert.isTrue(err instanceof Error);
+				assert.equal(err.message, 'abc already exists and no overwrite flag was passed');
+			},
+
+			'feature name is lowercased'() {
```
agubler, line 141: Do you also need a test that includes a test that when passing an uppercase string to `has` it lowercased?

agubler, line 141: I can see it now, you can ignore.

matt-gadd, line 141: thats covered above in the has test: https://github.com/dojo/core/pull/173/files#diff-9fb8a35085488b0cff49fd51446a500cR61

kitsonk:
We should do this on `dojo/has` now.

PR #178, opened 2016-07-19
Migration to TS2.0

### LINKED ISSUE #147: TypeScript 2.0 for dojo/core

agubler:
Dependant on https://github.com/dojo/shim/pull/2
- [x] Upgrade to TS2.0 (beta)
- [x] Add `strictNullChecks`
- [x] Add `noImplicitThis`

Fixes #147


src/DateObject.ts
```diff
@@ -36,7 +36,7 @@ export interface DateProperties {
 	year: number;
 }
 
-const days = [ null, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 ];
+const days = [ -1, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31 ];
```
kitsonk: Maybe we should add a comment here overall... because this is an odd array and has no documentation.  Would `NaN` be appropriate here instead of `-1`?

src/Registry.ts
```diff
@@ -32,11 +32,11 @@ export default class Registry<T> {
 	 * @returns the matching value, or a default value if one exists.
 	 */
 	match(...args: any[]): T {
-		let entries = this._entries.slice(0);
+		const entries = this._entries ? this._entries.slice(0) : [];
```
kitsonk, line 35: Not an issue with this change but seeing this we should be using `import { from as arrayFrom } from 'dojo-shim/array'` here...

src/Registry.ts
```diff
@@ -55,21 +55,23 @@ export default class Registry<T> {
...
+		let entry: Entry<T> | null = {
 			test: test,
 			value: value
 		};
 
 		(<any> entries)[(first ? 'unshift' : 'push')](entry);
 
 		return {
-			destroy: function () {
+			destroy: function (this: any) {
```
kitsonk: `this: Handle` should work here, right?

src/Scheduler.ts
```diff
@@ -3,7 +3,7 @@ import { QueueItem, queueTask } from './queue';
 
 function getQueueHandle(item: QueueItem): Handle {
 	return {
-		destroy: function () {
+		destroy: function (this: any) {
```
kitsonk: `this: Handle`

src/Registry.ts
agubler: yup, I think you are right.

src/UrlSearchParams.ts
```diff
@@ -115,19 +118,20 @@ export default class UrlSearchParams {
 	 * @param key The key to return the first value for
 	 * @return The first string value for the key
 	 */
-	get(key: string): string {
+	get(key: string): string | null | undefined {
```
kitsonk: Again, the change I don't argue with, but `null` seems inconsistent for `get()`...  I think people would expect `undefined`.

src/Registry.ts
agubler, line 35: Occurrences like this are probably something that we should review across the entire code base as a separate concern.

I can of course fix this single one :)

src/UrlSearchParams.ts
agubler: Perhaps that is more fluent for the API, however we were already returning `null` then `this.has(key)` is `false` - I can simply convert to use `undefined` in the instance?

agubler:
``` js
    /**
     * Returns the first value associated with a key.
     * @param key The key to return the first value for
     * @return The first string value for the key
     */
    get(key: string): string | undefined {
        if (!this.has(key)) {
            return undefined;
        }
        const value = this._list[key];
        return value ? value[0] : undefined;
    }
```

src/Registry.ts
agubler, line 35: Also the API for `from` in the `array` shim doesn't support `null`.

src/UrlSearchParams.ts
kitsonk: 👍

src/Registry.ts
kitsonk, line 35: BAH! Pedantic... 😛 dojo/meta#63 and we will clean up later.

src/DateObject.ts
agubler: 👍

src/streams/ReadableStreamReader.ts
```diff
@@ -201,11 +206,14 @@ export default class ReadableStreamReader<T> {
 	 */
 	resolveReadRequest(chunk: T): boolean {
 		if (this._readRequests.length > 0) {
-			this._readRequests.shift().resolve({
-				value: chunk,
-				done: false
-			});
-			return true;
+				const readRequest = this._readRequests.shift();
```
agubler: alignment

src/UrlSearchParams.ts
```diff
@@ -115,21 +118,22 @@ export default class UrlSearchParams {
 	 * @param key The key to return the first value for
 	 * @return The first string value for the key
 	 */
-	get(key: string): string {
+	get(key: string): string |  undefined {
```
kitsonk: There are two spaces here...  I know, being pedantic.

src/request.ts
```diff
@@ -9,7 +9,7 @@ import { ParamList } from './UrlSearchParams';
 declare var require: any;
 
 export class FilterRegistry extends Registry<RequestFilter> {
-	register(test: string | RegExp | RequestFilterTest, value: RequestFilter, first?: boolean): Handle {
+	register(test: string | RegExp | RequestFilterTest | null, value: RequestFilter, first?: boolean): Handle {
```
kitsonk, line 12: Why would we want to ever register a `null` test?

src/request.ts
```diff
@@ -19,7 +19,7 @@ export class FilterRegistry extends Registry<RequestFilter> {
 		}
 		else if (test instanceof RegExp) {
 			entryTest = (response, url, options) => {
-				return test.test(url);
+				return test ? (<RegExp> test).test(url) : null;
```
kitsonk: You could work around this guard being reset (and the subsequent need to cast) by doing something like this at the top of the function scope:

``` typescript
const inTest = test;
```

The TypeScript team are likely to add the ability to declare arguments as `const` though so that type narrowing doesn't get reset in function blocks like this.

kitsonk:
Also for the record, I am currently getting:

```
src/async/iteration.ts(169,43): error TS2345: Argument of type '(i?: number | undefined) => T | Thenable<T> | undefined' is not assignable to parameter of type '((value?: number | undefined) => T | Thenable<T>) | undefined'.
  Type '(i?: number | undefined) => T | Thenable<T> | undefined' is not assignable to type '(value?: number | undefined) => T | Thenable<T>'.
    Type 'T | Thenable<T> | undefined' is not assignable to type 'T | Thenable<T>'.
      Type 'undefined' is not assignable to type 'T | Thenable<T>'.
src/async/iteration.ts(202,15): error TS2345: Argument of type '(result: { values: T[] | undefined; results: U[] | undefined; } | undefined) => U[] | null | unde...' is not assignable to parameter of type '((value?: { values: T[] | undefined; results: U[] | undefined; } | undefined) => U[] | Thenable<U...'.
  Type '(result: { values: T[] | undefined; results: U[] | undefined; } | undefined) => U[] | null | unde...' is not assignable to type '(value?: { values: T[] | undefined; results: U[] | undefined; } | undefined) => U[] | Thenable<U[]>'.
    Type 'U[] | null | undefined' is not assignable to type 'U[] | Thenable<U[]>'.
      Type 'null' is not assignable to type 'U[] | Thenable<U[]>'.
```

When doing a `grunt` on this branch.

agubler:
@kitsonk if you remove the `typings` directory it should compile correctly, do you think we should clean the typings dir as part of the `dev` and `dist` tasks?

src/request.ts
agubler: This does prevent the need for a cast, do we think this is better than casting?

``` js
else if (test instanceof RegExp) {
  const inTest = test;
  entryTest = (response, url, options) => {
    return inTest ? inTest.test(url) : null;
  };
}
```

src/request.ts
agubler, line 12: This is due to the `Registry` class specifying `null` as the test here https://github.com/dojo/core/pull/178/files#diff-b0f17e2a57f905a1a051ae1f86ebc3cdR58 because it is set to `null` here https://github.com/dojo/core/pull/178/files#diff-b0f17e2a57f905a1a051ae1f86ebc3cdR76

kitsonk, line 12: ok, thanks

src/request.ts
kitsonk: I meant at the top of the function block, so it applies to the whole function scope (and might make it easier to eliminate when we have `const` params), but the code could almost revert back the original form, because `null` has also been eliminated from the type by the `instanceof` guard, right?

``` typescript
const inTest = test;
/* ... */
else if (inTest instanceof RegExp) {
  entryTest = (response, url, options) => inTest.test(url);
}
```

kitsonk:
Ah, ok, but as far as cleaning typings, see dojo/grunt-dojo2#4 which has been open for a while, because it isn't the first time it has happened... 😭

src/request.ts
agubler: gotcha

kitsonk:
👍

agubler:
There are some test failure, fixing.

agubler:
Unit tests failures fixed.

PR #185, opened 2016-08-19
bump to specific TS2 version and use tslint 'next' version

agubler:


src/async/Task.ts
```diff
@@ -10,7 +10,7 @@ export const Canceled = <State> 4;
  * @param value The value to guard
  */
 export function isTask<T>(value: any): value is Task<T> {
-	return Boolean(isThenable(value) && typeof value.cancel === 'function' && Array.isArray(value.children));
+	return Boolean(isThenable(value) && typeof (<any> value).cancel === 'function' && Array.isArray((<any> value).children));
```
agubler: open to suggestions on this one :)

kitsonk: I think the `isThenable` type guard is narrowing your `any` to a `Thenable`, so what you want to do is either reorder it or replicate the logic of the guard in `isThenable` without using it.

src/request.ts
```diff
@@ -196,8 +196,8 @@ export default request;
  * Add a filter that automatically parses incoming JSON responses.
  */
 filterRegistry.register(
-	function (response: Response<any>, url: string, options: RequestOptions) {
-		return typeof response.data && options && options.responseType === 'json';
+	function (response: Response<any>, url: string, options: RequestOptions): boolean {
+		return !!(typeof response.data) && options && options.responseType === 'json';
```
kitsonk: or `Boolean(typeof response.data && options && options.responseType === 'json');` which is a bit more clear than `!!`

agubler: roger that.

src/async/Task.ts
agubler: It was the `isThenable` narrowing the type, so something like this `return Boolean(value && typeof value.cancel === 'function' && Array.isArray(value.children) && isThenable(value));` works with out the type casting back to `any`

kitsonk:
👍

PR #186, opened 2016-08-22
try to get tests to run

agubler:


agubler:
yup this basic test matrix runs locally with 

```
grunt
grunt intern:node --combined
grunt intern:saucelabs --combined
```

agubler:
so why does the timeout value in `tests/unit/util.ts` affect other tests, such as `× async/timing - DelayedRejection - is eventually rejected with error (0.102s)` failure occurs when this timeout is increased.

agubler:
screen shot 2016-08-22 at 23 09 23

😢

PR #187, opened 2016-08-24
Fix fragile test cases

### LINKED ISSUE #107: Debounce/Throttle test cases too fragile
It seems that randomly we get failures to the [throttle and debounce test cases](https://travis-ci.org/dojo/core/builds/102064338#L174).  Usually Firefox but sometimes other platforms.

We should find a way to make them more robust.

kitsonk:
For some reasons, in certain runtime environments and async code it appears that sinon does not consitently track `callCount`.  Therefore sometimes the throttle tests would fail.  This PR rewrites these not to use the sinon spy.

In addition it lengthens the default timeout, because sometimes the functional tests for the text loader take longer during heavy traffic periods with either Travis or SauceLabs and delay the responses from the server, causing false failures.

Fixes #107


agubler:
👍

PR #189, opened 2016-09-05
Support applying aspects to MapLike objects

kitsonk:
`aspect` now supports applying advice to Map like objects (e.g. ones that utilise key value pairs using a `.set()` and `.get()` functionality).

This is to support allowing `Map`s to be used when keeping a map of listeners, thereby supporting dojo/compose#52.

Also, this PR cleans up `aspect` to align it to current coding standards, including using string literals, eliminates the "ugliness" of the `null` assignments when unassigning values, removing `<any>` casting and adding comments to provide clarity.

### DIFF
--- a/src/aspect.ts
+++ b/src/aspect.ts
@@ -1,27 +1,80 @@
 import { Handle } from './interfaces';
 import { createHandle } from './lang';
 
+/**
+ * An object that provides the necessary APIs to be MapLike
+ */
+export interface MapLike<K, V> {
+	get(key: K): V;
+	set(key: K, value?: V): this;
+}
+
+/**
+ * An internal type guard that determines if an value is MapLike or not
+ *
+ * @param value The value to guard against
+ */
+function isMapLike(value: any): value is MapLike<any, any> {
+	return value && typeof value.get === 'function' && typeof value.set === 'function';
+}
+
+export interface Indexable {
+	[method: string]: any;
+}
+
+/**
+ * The types of objects or maps where advice can be applied
+ */
+export type Targetable = MapLike<string, any> | Indexable;
+
+type AdviceType = 'before' | 'after';
+
+/**
+ * A meta data structure when applying advice
+ */
 interface Advised {
 	id?: number;
-	advice: Function | null;
-	previous?: Advised | null;
-	next?: Advised | null;
+	advice?: Function;
+	previous?: Advised;
+	next?: Advised;
 	receiveArguments?: boolean;
 }
 
+/**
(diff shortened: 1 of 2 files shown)


agubler:
👍

src/aspect.ts
```diff
@@ -1,27 +1,80 @@
...
+/**
+ * An internal type guard that determines if an value is MapLike or not
+ *
+ * @param value The value to guard against
+ */
+function isMapLike(value: any): value is MapLike<any, any> {
+	return value && typeof value.get === 'function' && typeof value.set === 'function';
+}
+
+export interface Indexable {
```
tomdye, line 21: I believe the exported interfaces should be in alphabetical order as per https://github.com/dojo/meta/blob/master/STYLE.md#ordering

kitsonk: I hadn't realised that was there...  The method I was using was keeping it close to the first use site, as I find it makes the code easier to read that way.

Do you feel ordering them alphabetically, when they are also used within the module, improves readability and maintainability, or were you just being pedantic?  If it is the former, then I would like to understand it better.  If it is the latter, it is easy to change the style guide.

rishson: Yep - don't care which we do, change code or guide, but lets aim to be examplars of our own standards.
In this instance, should we remove from the guide if we can't find violations with a linter?

I don't know if `tslint` can detect orderings, didn't last time I looked.

tomdye: I've been pulled me up for the same before in `dojo/loader`. I believe it's the interface names more than anything, ie. `MapLike` above `Indexable`, that's not the correct order.
We either need to code to our guidelines or update them.

kitsonk: We debated this before, regarding "should we have anything we can't `tslint`" (see: dojo/meta#43) and in my opinion, if that were true, then we would have unmaintainable code as there are always higher order and subjective things a linter can't enforce.

> We either need to code to our guidelines or update them.

I agree, I was trying to determine why you raised it to help determine if we needed to change them or not.  I guess it makes sense for exported interfaces to be alphabetically ordered at the top of the file, after the imports, where as other items, I still think make better sense close to their first usage site.

tomdye: I guess it was raised as an example of us not meeting our own guidelines, indicating that we should probably update them. Keeping exported interfaces at the top however is important, their order at the top, not so much.

agubler:
Okay can we take the styling offline? I think after that then this is good to go 👍

tomdye:
merge away 👍

PR #192, opened 2016-09-23
Improved the global object detection code

### LINKED ISSUE #182: Accessing global
It appears that in some environments, the global scope is not referencable via our current method:

``` typescript
const globalObject: any = Function('return this')();
```

In particular when some CSP restrictions are in place.  There is a TC39 proposal to make it available and accessible in a consistent place and describes [the issue](https://github.com/tc39/proposal-global#rationale).

We should consider changing this to:

``` typescript
const globalObject: any = function() {
    if (typeof self !== 'undefined') {
        return self;
    }
    if (typeof window !== 'undefined') {
        return window;
    }
    if (typeof global !== 'undefined') {
        return global;
    }
    throw new Error('unable to locate global object');
}();
```

bryanforbes:
Fixes #182

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -1,2 +1,17 @@
-const globalObject: any = Function('return this')();
+const globalObject: any = (function (): any {
+	if (typeof window !== 'undefined') {
+		// Browsers
+		return window;
+	}
+	else if (typeof global !== 'undefined') {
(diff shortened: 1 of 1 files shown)


kitsonk (approved):
👍

PR #193, opened 2016-09-26
Add IdentityRegistry and rename Registry to MatchRegistry

novemberborn:
Include `IdentityRegistry` from `dojo-app`, see https://github.com/dojo/app/issues/7.

**This renames `dojo-core/Registry` to `dojo-core/MatchRegistry`**. @kitsonk seemed OK with that back in June when I proposed the rename, but let me know if it's actually a concern.

I can submit PRs for any other repos that reference `dojo-core/Registry` once the rename is approved.

### DIFF
--- a/src/IdentityRegistry.ts
+++ b/src/IdentityRegistry.ts
@@ -1,0 +1,147 @@
+import Map from 'dojo-shim/Map';
+import WeakMap from 'dojo-shim/WeakMap';
+import { Handle } from './interfaces';
+
+const noop = () => {};
+
+interface Entry<V> {
+	handle: Handle;
+	value: V;
+}
+
+interface State<V> {
+	entryMap: Map<Identity, Entry<V>>;
+	idMap: WeakMap<V, Identity>;
+}
+
+const privateStateMap = new WeakMap<IdentityRegistry<any>, State<any>>();
+
+function getState<V>(instance: IdentityRegistry<V>): State<V> {
+	return privateStateMap.get(instance);
+}
+
+/**
+ * Registry identities can be strings or symbols. Note that the empty string is allowed.
+ */
+export type Identity = string | symbol;
+
+/**
+ * A registry of values, mapped by identities.
+ */
+export default class IdentityRegistry<V extends Object> {
+	constructor() {
+		privateStateMap.set(this, {
+			entryMap: new Map<Identity, Entry<V>>(),
+			idMap: new WeakMap<V, Identity>()
+		});
+	}
+
+	/**
+	 * Look up a value by its identifier.
+	 *
+	 * Throws if no value has been registered for the given identifier.
+	 *
+	 * @param id The identifier
+	 * @return The value
+	 */
+	get(id: Identity): V {
+		const entry = getState<V>(this).entryMap.get(id);
+		if (!entry) {
+			throw new Error(`Could not find a value for identity '${id.toString()}'`);
+		}
+
+		return entry.value;
+	}
+
+	/**
+	 * Determine whether the value has been registered.
+	 * @param value The value
+	 * @return `true` if the value has been registered, `false` otherwise
+	 */
+	contains(value: V): boolean {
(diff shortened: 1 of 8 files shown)


novemberborn:
Could consider aliasing `dojo-core/Registry` to make releasing easier.

src/IdentityRegistry.ts
```diff
@@ -0,0 +1,137 @@
...
+		entry.handle.destroy();
+		return true;
+	}
+
+	/**
+	 * Determine whether a value has been registered for the given identifier.
+	 * @param id The identifier
+	 * @return `true` if a value has been registered, `false` otherwise
+	 */
+	hasId(id: Identity): boolean {
```
kitsonk: Why not `has`?  I can guess you might say "well I am not passing the object" but that is similiar to a Map, in that you pass the [key](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has), not the object.

src/IdentityRegistry.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Registry identities can be strings or symbols. Note that the empty string is allowed.
+ */
+export type Identity = string | symbol;
+
+/**
+ * A registry of values, mapped by identities.
+ */
+export default class IdentityRegistry<V extends Object> {
+	private _entryMap: Map<Identity, Entry<V>>;
```
kitsonk: Can we stored `_entryMap` and `_idMap` in a WeakMap?

src/IdentityRegistry.ts
```diff
@@ -0,0 +1,137 @@
...
+		// Adding the same value with the same id is a noop, return the original handle.
+		if (existingEntry && existingId) {
+			return existingEntry.handle;
+		}
+
+		let handle = {
+			destroy: () => {
+				handle.destroy = noop;
+				this._entryMap.delete(id);
+				this._idMap.delete(value);
```
kitsonk: Deleting from a WeakMap! 😝

kitsonk (commented):
Also, while not critical, it would be great to add some iterator interfaces to this, like `[Symbol.iterator]`, `entries()`, `values()` and `ids()`.  I guess we could tackle that as a separate issue though.

novemberborn:
> Also, while not critical, it would be great to add some iterator interfaces to this, like [Symbol.iterator], entries(), values() and ids(). I guess we could tackle that as a separate issue though.

Yes.

novemberborn:
Updated.

src/IdentityRegistry.ts
```diff
@@ -18,12 +21,9 @@ export type Identity = string | symbol;
  * A registry of values, mapped by identities.
  */
 export default class IdentityRegistry<V extends Object> {
-	private _entryMap: Map<Identity, Entry<V>>;
-	private _idMap: WeakMap<V, Identity>;
-
 	constructor() {
-		this._entryMap = new Map<Identity, Entry<V>>();
-		this._idMap = new WeakMap<V, Identity>();
+		entryMaps.set(this, new Map<Identity, Entry<V>>());
```
kitsonk: Sorry, to make you do this again, but we have identified an ideal pattern... When we have a 1:1 relationship between private entries in a WeakMap, we should create an interface for managing them and store only one value.  See: dojo/meta#74.  (Or feel free to input on why it is a stupid pattern and we should do something else)

kitsonk (changes requested):
Sorry... I should have been more explicit the first time...

src/IdentityRegistry.ts
kitsonk: By the way, you even gave it a 👍 so you must like that pattern!

novemberborn: Oh yeah! I think I came up with it, no? 😉

novemberborn:
Updated again.

Used a helper method so the generic value type can be read out of the private state map.

kitsonk (approved):
👍

PR #194, opened 2016-09-26
Allow null and undefined sources in lang.assign()

novemberborn:
Object.assign() ignores null and undefined source values, do the same here.


kitsonk (approved):
👍

PR #196, opened 2016-09-29
clean-up readme language for beta

dylans:
meta says beta though we've not pushed to npm yet, so should land this when we tag for beta?


kitsonk (approved):
👍

Issue #197, opened 2016-10-02
Can't send any POST body with request/node


sebilasse:
in `request/node` : 

requestOptions SHOULD have at least `body` :

``` typescript
interface HttpsOptions extends Options {
    body?: any;
    // ...
}
```

and

``` typescript
const requestOptions: HttpsOptions = {
    agent: options.agent,
    auth: parsedUrl.auth || options.auth,
    body: options.data,
    // ...
}
```

additionally I think there are two more options when sending POST data in nodeJS :
`form: any;`
and
`json: boolean;`

dylans:
@sebilasse thanks, that seems reasonable. Do you intend to create a pull request, or should we handle that?

sebilasse:
@dylans Will investigate the next days (not sure about the `json` parameter, poor node documentation on that). Might be helpful to see how https://github.com/request/request deals with it. 
However, created PR :

sebilasse:
some more findings, all about `request/node`
It is also important to set the headers as lowercase 

``` js
if (typeof options.headers === 'object') {
    options.headers = Object.keys(options.headers).reduce(function(o, key) {
        o[key.toLowerCase()] = options.headers[key];
        return o;
    }, {});
}
options = mixin({maxRedirects: 5}, options);
options.headers = Object.keys(options.headers).reduce(function(o, key) {
    o[key.toLowerCase()] = options.headers[key];
    return o;
}, {});
```

if you e.g. just check 
`if (!('user-agent' in requestOptions.headers)) {`

What I also stumbled upon is this note in `request/node` :
_"TODO: This redirect code is not 100% correct according to the RFC; needs to handle redirect loops and restrict/modify certain redirects"_
This feels pretty important and could also be a (trackable) issue (especially because redirect loops will occur). Apart from that I guess that if an options.headers.host is already set it MUST be rewritten by the redirect code (think of `www.` redirects) :

``` js
    if (options.headers && options.headers['host'] && requestOptions.host) {
        options.headers.host = requestOptions['host'];
    }
```

My temporary fix to prevent infinite loops in node is to send a `redirectCount` with the options which can be limited by `maxRedirects` and never allows more than 5 –

for now see this [GIST](https://gist.github.com/sebilasse/e54fbd7722c98de5ad1e7dd8ab91f52d) for the complete file, modifications are marked by "TODO MOD"

rorticus:
[...]

PR #198, opened 2016-10-02
POST body with request/node

sebilasse:
see https://github.com/dojo/core/issues/197

### DIFF
--- a/src/request/node.ts
+++ b/src/request/node.ts
@@ -29,6 +29,7 @@ interface Options {
 }
 
 interface HttpsOptions extends Options {
+	body?: any;
 	ca?: any;
 	cert?: string;
 	ciphers?: string;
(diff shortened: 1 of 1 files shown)


kitsonk:
@sebilasse thanks for the contribution.  The PR looks fine, but it would be great to get tests for it before we were to merge it.

sebilasse:
What would a test look like without a server?

What I also stumbled upon is the note in `request/node`:
"TODO: This redirect code is not 100% correct according to the RFC; needs to handle redirect loops and restrict/modify certain redirects"
This feels pretty important and could also be a (trackable) issue (especially because redirect loops will occur, try german lead media site "spiegel.de" [without 'www.']).

kitsonk:
There is a server integrated into the test harness.  The server code is located in `tests/services/echo.ts` and is bootstrapped when the tests are run.

rorticus:
Closing because issue was closed.

PR #203, opened 2016-10-04
improve request/util destructuring

dylans:
Suggestion from Kit in reviewing conference talk materials, as a way to improve our usage of destructuring.

### DIFF
--- a/src/request/util.ts
+++ b/src/request/util.ts
@@ -9,8 +9,8 @@ import UrlSearchParams from '../UrlSearchParams';
  */
 export function generateRequestUrl(url: string,
 		{ query, cacheBust }: RequestOptions = {}): string {
+	query = new UrlSearchParams(query).toString();
 	if (cacheBust) {
-		query = new UrlSearchParams(query).toString();
 		const bustString = String(Date.now());
 		query += query ? `&${bustString}` : bustString;
 	}
--- a/src/request/util.ts
+++ b/src/request/util.ts
@@ -5,13 +5,12 @@ import UrlSearchParams from '../UrlSearchParams';
  * Returns a URL formatted with optional query string and cache-busting segments.
  *
  * @param url The base URL.
- * @param options The options hash that is used to generate the query string.
+ * @param options The RequestOptions used to generate the query string or cacheBust.
  */
 export function generateRequestUrl(url: string,
-		{ cacheBust }: RequestOptions = {}): string {
-	let query = '';
+		{ query, cacheBust }: RequestOptions = {}): string {
 	if (cacheBust) {
-		query = new UrlSearchParams(cacheBust.query).toString();
+		query = new UrlSearchParams(query).toString();
 		const bustString = String(Date.now());
 		query += query ? `&${bustString}` : bustString;
 	}
--- a/src/request/util.ts
+++ b/src/request/util.ts
@@ -9,8 +9,9 @@ import UrlSearchParams from '../UrlSearchParams';
  */
 export function generateRequestUrl(url: string,
 		{ cacheBust }: RequestOptions = {}): string {
-	let query = new UrlSearchParams(cacheBust.query).toString();
+	let query = '';
 	if (cacheBust) {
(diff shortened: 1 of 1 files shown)


kitsonk (changes requested):
tabs not spaces... should run `grunt` to do a dev build prior to submission 😛

kitsonk:
Ping...

src/request/util.ts
```diff
@@ -7,14 +7,13 @@ import UrlSearchParams from '../UrlSearchParams';
  * @param url The base URL.
  * @param options The options hash that is used to generate the query string.
  */
-export function generateRequestUrl(url: string, options: RequestOptions): string {
+export function generateRequestUrl(url: string,
+		{ cacheBust }: RequestOptions = {}): string {
```
kitsonk: `query` needs to be de-structured as well to avoid the issue in line 12 which the compiler is complaining about, because now there is no `options`

kitsonk (changes requested):
Needs a change to compile properly.

dylans:
@kitsonk should compile properly. Lessons learned:
- Don't copy an example from a conference slide where tabs are converted to spaces
- Avoid grunt/npm hell
- JSDoc doesn't have a proper way to document this syntax (that we know of)

src/request/util.ts
```diff
@@ -9,7 +9,7 @@ import UrlSearchParams from '../UrlSearchParams';
  */
 export function generateRequestUrl(url: string,
 		{ cacheBust }: RequestOptions = {}): string {
-	let query = new UrlSearchParams(options.query).toString();
+	let query = new UrlSearchParams(cacheBust.query).toString();
```
kitsonk: `cacheBust.query` is not a thing...

``` ts
export function generateRequestUrl(url: string,
        { cacheBust, query }: RequestOptions = {}): string {
    let query = new UrlSearchParams(query).toString();
```

kitsonk (changes requested):
...

src/request/util.ts
```diff
@@ -7,14 +7,13 @@ import UrlSearchParams from '../UrlSearchParams';
...
  */
-export function generateRequestUrl(url: string, options: RequestOptions): string {
-	let query = new UrlSearchParams(options.query).toString();
-
-	if (options.cacheBust) {
-		const cacheBust = String(Date.now());
-		query += query ? '&' + cacheBust : cacheBust;
+export function generateRequestUrl(url: string,
+		{ cacheBust }: RequestOptions = {}): string {
+	let query = new UrlSearchParams(cacheBust.query).toString();
```
kitsonk: `cacheBust.query` is not a thing...

``` ts
export function generateRequestUrl(url: string,
        { cacheBust, query }: RequestOptions = {}): string {
    let query = new UrlSearchParams(query).toString();
```

kitsonk (changes requested):
...

src/request/util.ts
```diff
@@ -5,16 +5,15 @@ import UrlSearchParams from '../UrlSearchParams';
...
-export function generateRequestUrl(url: string, options: RequestOptions): string {
-	let query = new UrlSearchParams(options.query).toString();
-
-	if (options.cacheBust) {
-		const cacheBust = String(Date.now());
-		query += query ? '&' + cacheBust : cacheBust;
+export function generateRequestUrl(url: string,
+		{ query, cacheBust }: RequestOptions = {}): string {
+	if (cacheBust) {
+		query = new UrlSearchParams(query).toString();
```
kitsonk: This needs to be moved outside of the `if (cacheBust)` to preserve the logic you are replacing

kitsonk (changes requested):
Change in logic from original.

dylans:
Thanks @kitsonk . It is passing CI, though failing on code coverage (which is unchanged from prior to this PR).

kitsonk (approved):

kitsonk:
Actually it is an issue with the way Istanbul is detecting/instrumenting the destructuring in the compiled code.  It is seeing some branches there and they don't get covered in the down emit.

PR #212, opened 2016-10-14
Making redirects follow (more closely anyway) the RFC, fixes #207

### LINKED ISSUE #207: Redirects and core/request
There is a `TODO` marked in the `core/request` code that indicates:

> This redirect code is not 100% correct according to the RFC; needs to handle redirect loops and restrict/modify certain redirects.

This needs to be addressed.

rorticus:
This should more closely follow an RFC.. I used this to find the requirements,

https://www.w3.org/Protocols/rfc2616/rfc2616.txt

In the places where the RFC is unclear, or where browsers (Chrome/Firefox) deliberately ignore the RFC, I tried to match them and call out in the code when that happens.

The RFC states that clients should "figure it out" when it comes to infinite redirect loops.  I wasn't sure if we wanted to get into that or not, so I'm just setting a (configurable) limit on the number of redirects we'll allow before we bail.

I also decided to include a `keepOriginalMethod` configuration to ignore the best practices and instead redirect how you _think_ it should redirect, i.e., `POST` requests redirect to `POST` requests.  I worked on an API once that had a very mean load balancer in front of it that would forward all my POST requests to another server with a 301 redirect. If the client I was working with didn't have a "keep original method" option, I would not have been able to interface with the API.

Anyways, let me know what you think!

### DIFF
--- a/src/request/node.ts
+++ b/src/request/node.ts
@@ -15,6 +15,8 @@ import { generateRequestUrl } from './util';
 // TODO: Where should the dojo version come from? It used to be kernel, but we don't have that.
 let version = '2.0.0-pre';
 
+const DEFAULT_REDIRECT_LIMIT = 15;
+
 interface Options {
 	agent?: any;
 	auth?: string;
@@ -62,6 +64,41 @@ export interface NodeRequestOptions<T> extends RequestOptions {
 	streamData?: boolean;
 	streamEncoding?: string;
 	streamTarget?: WritableStream<T>;
+	redirectOptions?: {
+		limit?: number;
+		count?: number;
+		keepOriginalMethod?: boolean;
+	};
+}
+
+function redirect<T>(resolve: (p?: any) => void, reject: (_?: Error) => void, url: string, options: NodeRequestOptions<T>): boolean {
+	if (!options.redirectOptions) {
+		options.redirectOptions = {};
+	}
+
+	const { limit: redirectLimit = DEFAULT_REDIRECT_LIMIT } = options.redirectOptions;
+	const { count: redirectCount = 0 } = options.redirectOptions;
+	const { followRedirects = true } = options;
+
+	if (!followRedirects) {
+		return false;
+	}
+
+	if (!url) {
+		reject(new Error('asked to redirect but no location header was found'));
+		return true;
+	}
+
+	if (redirectCount > redirectLimit) {
+		reject(new Error(`too many redirects, limit reached at ${redirectLimit}`));
+		return true;
+	}
+
+	options.redirectOptions.count = redirectCount + 1;
+
+	resolve(node(url, options));
+
+	return true;
 }
 
 export default function node<T>(url: string, options: NodeRequestOptions<T> = {}): ResponsePromise<T> {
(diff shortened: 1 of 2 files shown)


novemberborn:
> In the places where the RFC is unclear, or where browsers (Chrome/Firefox) deliberately ignore the RFC, I tried to match them and call out in the code when that happens.

RFC 2616 has been obsoleted in favor of https://tools.ietf.org/html/rfc7230. Not sure if anything is more clear in the newer spec though.

rorticus:
> RFC 2616 has been obsoleted

Thanks, I'll check it out and see if anything is different. I'm not sure whats up with the CI stuff, I'm pretty sure all the tests pass locally, but I'll take a look at that too.

kitsonk:
I wonder if there was a regression/change in NodeJS v6.8.0 (which is what Travis is using).

rorticus:
> RFC 2616 has been obsoleted

OK so RFC 7230 is remarkably silent on redirects (redirects are only mentioned in passing, when referring to how they affect the actual http protocol).

rorticus:
> I wonder if there was a regression/change in NodeJS v6.8.0 (which is what Travis is using).

I don't know.. I just ran the tests locally with node 6.8.1 and they all passed.

kitsonk:
I restarted the build on Travis to see if was just in Travis mind... ;-)

src/request/node.ts
```diff
@@ -62,6 +64,41 @@ export interface NodeRequestOptions<T> extends RequestOptions {
...
+	};
+}
+
+function redirect<T>(resolve: (p?: any) => void, reject: (_?: Error) => void, url: string, options: NodeRequestOptions<T>): boolean {
+	if (!options.redirectOptions) {
+		options.redirectOptions = {};
+	}
+
+	const redirectLimit = options.redirectOptions.limit || defaultRedirectLimit,
+		{ count: redirectCount = 0 } = options.redirectOptions,
```
kitsonk: [Style Guide, Variables Item 3](https://github.com/dojo/meta/blob/master/STYLE.md#variables) indicates that every variable should be declared on its own line with it own variable keyword, instead of using a list.

src/request/node.ts
```diff
@@ -62,6 +64,41 @@ export interface NodeRequestOptions<T> extends RequestOptions {
...
+		keepOriginalMethod?: boolean;
+	};
+}
+
+function redirect<T>(resolve: (p?: any) => void, reject: (_?: Error) => void, url: string, options: NodeRequestOptions<T>): boolean {
+	if (!options.redirectOptions) {
+		options.redirectOptions = {};
+	}
+
+	const redirectLimit = options.redirectOptions.limit || defaultRedirectLimit,
```
kitsonk: You could destructure this with a default value like the next line.

src/request/node.ts
```diff
@@ -62,6 +64,41 @@ export interface NodeRequestOptions<T> extends RequestOptions {
...
+		{ count: redirectCount = 0 } = options.redirectOptions,
+		{ followRedirects = true } = options;
+
+	if (!followRedirects) {
```
kitsonk, line 83: You could short circuit this function far earlier by moving this block further up, right after you pluck the `followRedirects` from the `options`.

tests/unit/request/node.ts
```diff
@@ -1,23 +1,193 @@
...
+			let error: any = null;
+
+			return nodeRequest(getRequestUrl(details.url), {
+				responseType: 'json',
+				method: method,
+				followRedirects: followRedirects,
+				redirectOptions: {
+					keepOriginalMethod
+				}
+			}).then((response) => {
```
kitsonk: We need to be more explicit about this in the style guide, but chains like this should linebreak before the next method... So:

```
})
.then((response) => {
```

kitsonk (changes requested):
A few thoughts/comments.

src/request/node.ts
rorticus: I don't understand... isn't this already the first thing that happens after we extract `followRedirects`?

src/request/node.ts
rorticus: Good catch! done.

src/request/node.ts
rorticus: Thanks for pointing that out. I think the last couple of projects I worked on had the _opposite_ rule :)

tests/unit/request/node.ts
rorticus: Fixed 👍

src/request/node.ts
```diff
@@ -15,6 +15,8 @@ import { generateRequestUrl } from './util';
 // TODO: Where should the dojo version come from? It used to be kernel, but we don't have that.
 let version = '2.0.0-pre';
 
+const defaultRedirectLimit = 15;
```
kitsonk: We should UPPER_CASE `const` values that we use as _constants_ like this.  So `DEFAULT_REDIRECT_LIMIT`

src/request/node.ts
```diff
@@ -144,17 +181,93 @@ export default function node<T>(url: string, options: NodeRequestOptions<T> = {}
...
-				response.statusCode < 400 &&
-				response.statusCode !== 304 &&
-				options.followRedirects !== false &&
-				nativeResponse.headers.location
+				response.statusCode < 400
 			) {
-				resolve(node(nativeResponse.headers.location, options));
-				return;
+				const redirectOptions = options.redirectOptions || {};
+				let newOptions = Object.create(options);
```
kitsonk: I can't see where `newOptions` gets reassigned.  If it doesn't, it should be a `const`.

kitsonk (commented):
Couple more comments

src/request/node.ts
rorticus: Updated!

src/request/node.ts
rorticus: Updated!

kitsonk (approved):

Issue #215, opened 2016-10-18
Add the ability to delay executing a function.


lzhoucs:
`async/timing.delay()` currently assumes a value will be returned. In some use cases, the user may want to delay executing a function that either returns a value or another promise. It would be nice to add this ability in `async/timing.delay()`

PR #216, opened 2016-10-18
Add the ability to delay executing a function. (#215)

lzhoucs:
This address #215.
Any feedback or comments are welcome.

### DIFF
--- a/src/async/timing.ts
+++ b/src/async/timing.ts
@@ -8,7 +8,7 @@ import { Thenable } from 'dojo-shim/interfaces';
  * @return {function (value: T | (() => T | Thenable<T>)): Promise<T>} a function producing a promise that eventually returns the value or executes the value function passed to it; usable with Thenable.then()
  */
 export function delay<T>(milliseconds: number): Identity<T> {
-	return function (value: T | (() => T | Thenable<T>)): Promise<T> {
+	return function (value?: T | (() => T | Thenable<T>)): Promise<T> {
 		return new Promise(function (resolve) {
 			setTimeout(function () {
 				resolve(typeof value === 'function' ? value() : value);
@@ -18,7 +18,7 @@ export function delay<T>(milliseconds: number): Identity<T> {
 }
 
 export interface Identity<T> {
-	(value: T | (() => T | Thenable<T>)): Promise<T>;
+	(value?: T | (() => T | Thenable<T>)): Promise<T>;
 }
 
 /**
--- a/src/async/timing.ts
+++ b/src/async/timing.ts
@@ -4,13 +4,13 @@ import Promise from 'dojo-shim/Promise';
  * Used for delaying a Promise chain for a specific number of milliseconds.
  *
  * @param milliseconds the number of milliseconds to delay
- * @return {function(T): Promise<T>} a function producing a promise that eventually returns the value passed to it; usable with Thenable.then()
+ * @return {function(T): Promise<T>} a function producing a promise that eventually returns the value or execute the value function passed to it; usable with Thenable.then()
  */
 export function delay<T>(milliseconds: number): Identity<T> {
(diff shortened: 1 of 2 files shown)


kitsonk:
We should re-base this and then in theory this should clear the CI error.

src/async/timing.ts
```diff
@@ -4,13 +4,13 @@ import Promise from 'dojo-shim/Promise';
  * Used for delaying a Promise chain for a specific number of milliseconds.
  *
  * @param milliseconds the number of milliseconds to delay
- * @return {function(T): Promise<T>} a function producing a promise that eventually returns the value passed to it; usable with Thenable.then()
+ * @return {function(T): Promise<T>} a function producing a promise that eventually returns the value or execute the value function passed to it; usable with Thenable.then()
```
kitsonk: `or executes the`

src/async/timing.ts
```diff
@@ -4,13 +4,13 @@ import Promise from 'dojo-shim/Promise';
...
  */
 export function delay<T>(milliseconds: number): Identity<T> {
 	return function (value: T): Promise<T> {
```
kitsonk: Actually, we should type this better, as it doesn't represent what will be returned properly.

It should be `return function (value: T | () => T | Thenable<T>): Promise<T>` which also likely means that I bet `Identity` also needs a change.

I would check the intellisense to validate that what you expect the `Promise<T>` to be is what it actually is, because I think you might have an issue.  Also, I think we should "unwrap" thenables too.

kitsonk (changes requested):
A couple of thoughts.

src/async/timing.ts
lzhoucs: I agree with unwrapping the `thenable`. However I think there might be a use case for it:

```
            const delayed = timing.delay(250);
            delayed('foo');
            delayed(function() { return 'bar'; });
```

Calling `delay` with a time but without a value, can sort of be used as a partial function. And the user can then apply that function multiple times without recreating the delay.  If it doesn't make sense, I can remove the `thenable` wrap.

lzhoucs:
I couldn't reproduce the following test failure in [build #317](https://travis-ci.org/dojo/core/builds/170904559) on Windows 7 IE 11.
"internet explorer 11.0 on Windows 7 - unit tests - async/timing - DelayedRejection - is eventually rejected with error"

src/async/timing.ts
```diff
@@ -1,23 +1,24 @@
...
 /**
  * Used for delaying a Promise chain for a specific number of milliseconds.
  *
  * @param milliseconds the number of milliseconds to delay
- * @return {function(T): Promise<T>} a function producing a promise that eventually returns the value passed to it; usable with Thenable.then()
+ * @return {function (value: T | (() => T | Thenable<T>)): Promise<T>} a function producing a promise that eventually returns the value or executes the value function passed to it; usable with Thenable.then()
  */
 export function delay<T>(milliseconds: number): Identity<T> {
-	return function (value: T): Promise<T> {
+	return function (value: T | (() => T | Thenable<T>)): Promise<T> {
```
agubler: Can we make the `value` optional? There could be use cases that don't care about the resolve value and simply need to create a delay.

kitsonk: I saw that in when you used it, I had the same thought, as long as we are in here. 👍

kitsonk:
@lzhoucs it seems to have been a transitory issue with SauceLabs.  Rerunning the CI came back clean.

kitsonk (approved):

src/async/timing.ts
lzhoucs: Thanks. That makes sense! I just pushed the updates.

kitsonk (approved):
👍

Issue #21, opened 2015-05-23
dojo-core/queue does not have default export


kitsonk:
Therefore, when importing you cannot do:

``` ts
import queueMicroTask from 'dojo-core/queue';
```

You can only:

``` ts
import queue = require('dojo-core/queue');
```

mwistrand:
You should be able to import as
`import { queueMicroTask } from './queue';`

PR #219, opened 2016-10-24
Refactored for removal of Promise shim from dojo/shim, shim issue #21

rorticus:
This is part of shim #21!  We removed the promise shim and are returning the native promise object.

It looks like in Typescript, if you cannot extend Promise and compile down to ES5. So in the mean time, I've created a Promise wrapper thingy, similar to what was in shim. The wrapper includes a `finally` because Task needed it.

It turns out that the promise shim from dojo/shim was swallowing uncaught errors. When you run the tests now, you get a ton of uncaught error messages (but the tests pass).

Did I mention this is dependent on https://github.com/dojo/shim/pull/41 ?

### DIFF
--- a/src/async/ExtensiblePromise.ts
+++ b/src/async/ExtensiblePromise.ts
@@ -1,0 +1,141 @@
+import { Iterable, forOf } from 'dojo-shim/iterator';
+import Promise, { Executor } from 'dojo-shim/Promise';
+import { Thenable } from 'dojo-shim/interfaces';
+
+/**
+ * Take a list of values, and if any are ExtensiblePromise objects, insert the wrapped Promise in its place,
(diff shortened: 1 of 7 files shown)


kitsonk:
We should rebase this off of master while we are getting dojo/shim#41 sorted out as there has been quite a bit of machinations in there with the integration of `dojo-interfaces`.

kitsonk (approved):
👍 🎃

kitsonk:
Actually this is now independent of dojo/shim#41, because it has now abstracted itself from that, which is good.  While there is a risk that when dojo/shim#41 gets published, it causes a regression, but both @rorticus and myself have tested against linked versions of this.  This is a great example of decoupling!

PR #221, opened 2016-10-25
Update dependencies and migrate to @types

kitsonk:
Refs dojo/meta#62

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,10 @@
   "name": "dojo-core",
   "version": "2.0.0-pre",
   "description": "Basic utilites for common TypeScript development",
+  "engines": {
+    "node": "^6.5.0",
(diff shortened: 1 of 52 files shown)


package.json
```diff
@@ -2,6 +2,10 @@
   "name": "dojo-core",
   "version": "2.0.0-pre",
   "description": "Basic utilites for common TypeScript development",
+  "engines": {
+    "node": "^6.5.0",
```
agubler: do you think we should just put 6.0.0 (which is the minimum version)? I am running `6.3.0`

agubler (approved):
one question

package.json
kitsonk: Adjusted, plus also the semver for the `@types` packages.

PR #230, opened 2016-11-02
Calling finally task for canceled tasks, fixes #205

### LINKED ISSUE #205: Task's finally callback not invoked if added after cancelation
The following test for `Task` should pass, but times out instead:

```
        'invoked if already canceled'(this: any) {
            const dfd = this.async();
            const task = new Task(() => {});
            task.cancel();

            task.finally(dfd.callback(() => {}));
        },
```

I think it's because `Task#finally()` defers to `Promise#finally()`, which in turn calls `Task#then()`, which doesn't invoke callbacks if the task is canceled.

rorticus:
If you call `finally` on an already canceled task, the `finally` task now runs.


kitsonk (approved):
👍

PR #231, opened 2016-11-03
No longer use conditional promise require in load

### LINKED ISSUE #224: load is incorrectly dynamically loading Promise
In ['src/load'](https://github.com/dojo/core/blame/master/src/load.ts#L16), Promise is being mis-loaded under NodeJS.  It isn't affecting anything because there is a native Promise in the versions of NodeJS we are using, but still it is a defect.

matt-gadd:
Fixes #224 

We should no longer need to perform this logic at all now.


kitsonk (approved):
👍 🦃

PR #232, opened 2016-11-04
Fixes to get successful transpilation and test runs on TS 2.1

rorticus:
Fixing some code in `Task` that causes incompatibilities with TypeScript 2.1

### DIFF
--- a/src/async/Task.ts
+++ b/src/async/Task.ts
@@ -80,24 +80,13 @@ export default class Task<T> extends ExtensiblePromise<T> {
 	 *
 	 */
 	constructor(executor: Executor<T>, canceler?: () => void) {
+		// we have to initialize these to avoid a compiler error of using them before they are initialized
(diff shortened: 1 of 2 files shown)


src/async/Task.ts
```diff
@@ -80,35 +80,47 @@ export default class Task<T> extends ExtensiblePromise<T> {
...
-		this._state = State.Pending;
-
-		this.children = [];
-		this.canceler = () => {
-			if (canceler) {
-				canceler();
-			}
-			this._cancel();
-		};
+		} catch (reason) {
```
kitsonk:
``` ts
}
catch (reason) {
```

kitsonk (commented):
One minor style comment

kitsonk:
Refs #238

src/async/Task.ts
rorticus: Good `catch`. Ha! Get it!?

PR #233, opened 2016-11-07
Comparison API

### LINKED ISSUE #138: Comparison API
Several functional problems have highlighted that `core` could benefit from a centralised API for comparing objects and mutating objects based upon those comparisons.  We should have an API that:
- Compares two objects/arrays/iterables? and provides a diff between the two, which would be a simple JavaScript object which then can be easily serialised and persisted.
- Take a set of comparison records and applies/patches them to an object.

kitsonk:
This PR implements a comparison API as part of core.

Fixes #138 

The Comparison API can deal with Arrays and "plain" Objects.  Plain objects are objects that have either `Object` as a constructor (or no constructor) and the properties values of the object are either primitives, arrays, or plain objects.  For objects, only enumerable own properties are considered.

The module `compare` exports two functions `diff` and `patch`.

`diff()` returns the difference between the first and the second argument and describes the steps to take to make the first object look like the second object.

`patch()` takes a target and applies a set of changes which will change the target.


kitsonk:
@maier49 @rorticus @vansimke could you please review

agubler:
@maier49 can you give this a review please?

maier49 (commented):
My main concern here is how this fits in with the existing comparison API provided in Stores. Stores uses a [patch implementation](https://github.com/dojo/stores/blob/master/src/patch/createPatch.ts#L4-L8) that serializes to [JSON Patch](https://tools.ietf.org/html/rfc6902).

I think the checks for plain objects and some other things in here are a little better than how Stores is currently handling it, but I don't know if we should replace the diff functionality without answering the question of how we want to serialize patches for sending to the server. If we do want to use JSON Patch, then it seems odd to switch to a patch/diff implementation that doesn't support that from one that does.

dylans:
We should probably decide what to do with this... do we need to leverage more of what dojo/stores has instead, etc.? @maier49 @agubler

maier49:
The [patch implementation in stores](https://github.com/dojo/stores/tree/master/src/patch) could pretty easily be moved over to core. It only relies on a few utility functions outside of the `patch` package. Since it already supports serialization to JSON patch and there's a PR open to support [JSON Merge Patch](https://github.com/dojo/stores/pull/80/files#diff-1944ded907c55ca34eed1feb7b57feb1) I'm still inclined to think that it would be better to move it over to core than to have competing implementations or switch to an implementation that doesn't yet have a plan for serialization.

We could address any issues with the API in the process of moving it over.

dylans:
@maier49 ok, let's do that!

agubler:
So shall we close this pending @maier49's work to move the implementation over from stores?

kitsonk:
I have refreshed this PR from what I have had to do in `intern-helper` which needed to use this API.

Also, my understanding is that @pottedmeat had a chat with @maier49 about the usage in stores for dgrid and they agreed that it might be better to take what is here and iterate on it to make it work for stores.  If that is still the case, we should land this and figure out if there are any features that are needed for stores.

Also, `widget-core` uses a `diff` function for properties, which we should evaluate if this API is sufficient for that, or if there are other changes required.

maier49:
@kitsonk the main thing for stores is just providing an appropriately serialized version of a diff for a `PATCH` request. I'm ok with `stores` using its own implementation until we figure that out here.

kitsonk:
I will take a look at what is in stores (and sorry again for not looking first) and see what it would take to do that.

src/compare.ts
```diff
@@ -0,0 +1,694 @@
...
+	const value = new Ctor(...(args || []));
+	if (propertyRecords) {
+		propertyRecords.forEach((record) => isConstructRecord(record) ? patchConstruct(value, record) : patchPatch(value, record));
+	}
+	defineProperty(target, name, assign({ value }, descriptor));
+	return target;
+}
+
+/**
+ * An internal function that take a value from array being patched and the target value from the same
```
rorticus: ~take~ takes

src/compare.ts
```diff
@@ -0,0 +1,694 @@
...
+				isArray(targetValue) ?
+					targetValue : [] : isPlainObject(targetValue) ?
+						targetValue : objectCreate(null),
+			patchValue
+		) :
+		patchValue;
+}
+
+/**
+ * Compares to plain objects or arrays and return a set of records which describe the differences between the two
```
rorticus: ~to~ two

rorticus (approved):
Just a few typos

Issue #229, opened 2016-11-01
Unhandled Promise Rejection Warnings


kitsonk:
Now that we have migrated to native Promises in dojo-shim, we are getting a [significant amount of unhandled promise rejection notifications](https://travis-ci.org/dojo/core/builds/171948284) in CI now that we should investigate and eliminate:

```
✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓(node:3631) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 21): undefined
✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓(node:3631) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 23): undefined
✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓✓DEPRECATED: This function will be removed in future versions.
✓DEPRECATED: foo
✓DEPRECATED: foo: This function will be removed in future versions.
✓DEPRECATED: This function will be removed in future versions.
    See foo for more details.
✓✓DEPRECATED: This function will be removed in future versions.
✓✓DEPRECATED: Foo#method: This function will be removed in future versions.
✓✓DEPRECATED: This function will be removed in future versions.
[...]

PR #235, opened 2016-11-10
Resolving unhandled promise rejection errors, #229

rorticus:
> The tests, they did throw uncaught promises a lot.
> Dispatching their promises, the results were forgot.
> Merge this PR,
> and then you will see,
> The test logs don't error, they're clean as can be.

Also, we'll probably want someone familiar with the streams to give this a quick glance to confirm that I'm not hiding any _important_ errors. There shouldn't be anything caught now that wasn't before, but, you never know!

### DIFF
--- a/src/request/node.ts
+++ b/src/request/node.ts
@@ -282,7 +282,8 @@ export default function node<T>(url: string, options: NodeRequestOptions<T> = {}
 						},
 						function (error: RequestError<T>) {
 							if (options.streamTarget) {
-								options.streamTarget.abort(error);
+								options.streamTarget.abort(error).catch(() => {
+								});
 							}
 							request.abort();
 							error.response = response;
@@ -318,7 +319,8 @@ export default function node<T>(url: string, options: NodeRequestOptions<T> = {}
 					resolve(response);
 				}
 				else {
-					options.streamTarget.close();
+					options.streamTarget.close().catch(() => {
+					});
 				}
 			});
 		});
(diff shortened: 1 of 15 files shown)


src/request/node.ts
```diff
@@ -282,7 +282,8 @@ export default function node<T>(url: string, options: NodeRequestOptions<T> = {}
 						},
 						function (error: RequestError<T>) {
 							if (options.streamTarget) {
-								options.streamTarget.abort(error);
+								options.streamTarget.abort(error).catch(() => {
```
kitsonk, line 287: It would be good to add a comment on this, because they look "odd" otherwise.

tests/unit/request/node.ts
```diff
@@ -122,6 +122,12 @@ const responseData: { [url: string]: DummyResponse } = {
 	}
 };
 
+process.on('unhandledRejection', (reason: any, promise: any) => {
```
kitsonk: Is this still desired/intended?  Seems like it was what was used to chase down the issues.

kitsonk (commented):
Just a couple of comments

tests/unit/request/node.ts
rorticus: Good catch, you're right, that shouldn't be there! :)

src/request/node.ts
rorticus: Do you think I should comment on all the places I did this, or just this one?

kitsonk (approved):
👍

PR #236, opened 2016-11-10
Case insensitivity when checking for user-agent header, fixes #168

### LINKED ISSUE #168: request: Header fields are case insensitive
see request/node, line 91:
`if (!('user-agent' in requestOptions.headers))`

I noticed two things:
- request/xhr does not set `user-agent`
- header fields are case insensitive,
  so what if ('User-Agent' in requestOptions.headers) ?

rorticus:
> `user-agent` checking was once case sensitive,
> sending it mixed case caused behavior preventative.
> now they are normalized,
> so send them all capitalized,
> and no longer will `USER-AGENT` be decorative.

Note that I although I changed some case sensitivity header checking in `request/xhr`, I did not do anything with setting the user agent.

See, 

https://dvcs.w3.org/hg/xhr/raw-file/tip/Overview.html#dom-xmlhttprequest-setrequestheader

### DIFF
--- a/src/request/node.ts
+++ b/src/request/node.ts
@@ -126,7 +126,7 @@ export default function node<T>(url: string, options: NodeRequestOptions<T> = {}
 
 	requestOptions.headers = options.headers || {};
 
-	if (!('user-agent' in requestOptions.headers)) {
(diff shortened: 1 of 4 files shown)


kitsonk (approved):
👍

Issue #217, opened 2016-10-24
Proposal: List


kitsonk:
For API consistency, it would be great if we had a `List`, which would essentially be an `Array`, but with an API that is consistent with `Map` and `Set`, so that operations could be performed on it without having to bifurcate the code path to deal with the collection.

The interface should be something like:

``` typescript
interface List<T> {
    readonly size: number;

    add(value: T): this;
    clear(): void;
    delete(idx: number): boolean;
    entries(): Iterator<[ number, T ]>;
    forEach(fn: (value: T, idx: number, list: this) => void, thisArg?: any): void;
    has(idx: number): boolean;
    keys(): Iterator<number>;
    values(): Iterator<T>;
    [Symbol.iterator]: Iterator<[ number, T ]>;

    includes(value: T): boolean;
    indexOf(value: T): number;
    lastIndexOf(value: T): number;
}
```

Are there other array extras we should have, like `reduce` and `reduceLeft`, `some`, `every`, etc?

kitsonk:
Oh and `push`, `pop`, `splice`, `join`?  The constructor should take and Iterable like other collections as well.

PR #237, opened 2016-11-14
Adding List, issue #217

rorticus:
Adding List, based on proposal in issue #217 .

### DIFF
--- a/src/List.ts
+++ b/src/List.ts
@@ -1,26 +1,6 @@
-import { Iterable, IterableIterator, ShimIterator } from 'dojo-shim/iterator';
+import { forOf, Iterable, IterableIterator, ShimIterator } from 'dojo-shim/iterator';
 
-export interface List<T> extends Iterable<T> {
-	readonly size: number;
-
-	add(value: T): this;
-	clear(): void;
-	delete(idx: number): boolean;
-	entries(): IterableIterator<[ number, T ]>;
-	forEach(fn: (value: T, idx: number, list: this) => void, thisArg?: any): void;
-	has(idx: number): boolean;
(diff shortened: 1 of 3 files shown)


src/List.ts
```diff
@@ -0,0 +1,106 @@
...
+	keys(): IterableIterator<number>;
+	lastIndexOf(value: T): number;
+	pop(): T | undefined;
+	push(value: T): void;
+	splice(start: number, deleteCount?: number, ...newItems: T[]): T[];
+	values(): IterableIterator<T>;
+}
+
+export default class ArrayList<T> implements List<T> {
+	private _items: T[];
```
kitsonk: I think we need to avoid private members on classes.  First, it isn't truly private, second, the `private` keyword won't be the way it gets implemented in ECMAScript and third, it causes all sorts of assignability issues when used in TypeScript.  Even though it is sort of "meta" I would use a weak map to store the items.  This is how they are suggesting people will eventually polyfill the private class members anyways.

src/List.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	[Symbol.iterator]() {
+		return this.values();
+	}
+
+	get size(): number {
+		return this._items.length;
+	}
+
+	constructor() {
```
kitsonk: Shouldn't this be able to take an argument of an interable as a source, like the other iterables?  So you can do this:

``` ts
new List([ 1, 2, 3 ]);
```

kitsonk (commented):
Initial review and a couple of comments

kitsonk (approved):
👍 awesomeness

Issue #225, opened 2016-10-28
Split streams into separate package


dylans:
Per the discussion in #90, we're going to move streams into a separate package given its size, as well as our usage for Observables over Streams for non-IO tracking features.

rorticus:
streams have been taken out of `dojo/core` and put into `dojo/streams`, but are stilled pulled in for use in `request/node`.  The next step is going to be to remove that dependency somehow. I talked with @matt-gadd and @agubler about it today and they are going to put in a related issue that hopefully can be used to resolve this one as well.

rorticus:
Split this up into https://github.com/dojo/core/issues/253 and https://github.com/dojo/core/issues/252.

PR #239, opened 2016-11-16
Removing streams and importing from dojo-streams, issue #225

rorticus:
Moving streams into a separate package,

https://github.com/rorticus/streams

I'm not really sure how to make `rorticus/streams` into `dojo/streams` but I'm guessing someone smart can tell me!


dylans (approved):
This is fine as a first pass, though I believe we were hoping to also refactor request/node.ts to not depend on streams so that this package wouldn't just pull in streams as a dependency. That said, we can do that refactoring/rework after we land this patch (which just moves streams into a separate package).

dylans:
> I'm not really sure how to make rorticus/streams into dojo/streams but I'm guessing someone smart can tell me!

Basically someone with admin in both accounts can move a repo, and then you can refork it once it has been moved.

rorticus:
> I believe we were hoping to also refactor request/node.ts to not depend on streams

I can try that! Would that be using observables instead?

dylans:
> Would that be using observables instead?

I'm not sure I know the right answer to that. To be fair, streams within Node.js usage is probably a good thing, but it feels wrong for all of core to rely on streams with the only dependency being the Node.js request API. Perhaps take a look at how it's being used and make a suggestion?

rorticus:
OK, so... it looks like it's _not_ being used... I [did a search](https://github.com/search?utf8=%E2%9C%93&q=org%3Adojo+user%3Adojo+streamTarget+in%3Afile&type=Code&ref=searchresults) for `streamTarget` (the name of the option that gets passed to a node request to use a stream)  on all the `dojo` repositories and the only hits are in the node request file...

dylans:
Right, the only thing that would eventually use it would be a Node.js request (which was to be supported for crypto and io, which have been deferred as we've not been focusing on Node.js APIs, though we want what we have to work inside Node.js... hopefully that distinction makes sense).

Issue #211, opened 2016-10-14
Support some form of Promise.all(Object)


kitsonk:
The ES6 specification only accepts iterators for an argument to `Promise.all()`.

In Dojo 1, our promise implementation used to accept either an `Array` or an `Object` map.  When dealing with a complex promise situation, it is very handy to have a _named_ result set, versus just an index based one.

@sebilasse requested we consider adding it to the `Promise` implementation in `dojo-shim` (see: dojo/shim#38) but that would mis-align the `Promise` implementation.  We do have `Task` where we could put that, but since `.all()` is a static method, we could also create a `promise.all` function that also accepted an iterator or an object map.

PR #241, opened 2016-11-17
Adding object support to ExtensiblePromise.all, issue #211

rorticus:
Allowing object support in `ExtensiblePromise.all` and it's derivatives (`Task.all`) (issue #211 ).

```ts
return ExtensiblePromise.all({
    one: 1, two: Promise.resolve(2), three: new ExtensiblePromise((resolve) => {
        setTimeout(() => resolve(3), 100);
    })
}).then((results: any) => {
    // results = { one: 1, two: 2, three: 3 }
});
```

Note that for this to work in IE 10/11, we need this... 

https://github.com/dojo/shim/pull/53

### DIFF
--- a/src/async/ExtensiblePromise.ts
+++ b/src/async/ExtensiblePromise.ts
@@ -17,6 +17,9 @@ function unwrapPromises(iterable: Iterable<any> | any[]): any[] {
 	return unwrapped;
 }
 
+export type DictionaryOfPromises<T> = { [_: string]: T | Promise<T> | Thenable<T> };
+export type ListOfPromises<T> = Iterable<(T | Thenable<T>)> | (T | Thenable<T>);
+
 /**
  * An extensible base to allow Promises to be extended in ES5. This class basically wraps a native Promise object,
  * giving an API like a native promise.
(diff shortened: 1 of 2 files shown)


src/async/ExtensiblePromise.ts
```diff
@@ -46,12 +46,33 @@ export default class ExtensiblePromise<T> {
...
+	 * ExtensiblePromise.all({ one: 1, two: 2 }).then(results => console.log(results));
+	 * // { one: 1, two: 2 }
+	 *
+	 * @param iterable    An iterable of values to resolve, or a key/value pair of values to resolve. These can be Promises, ExtensiblePromises, or other objects
 	 * @returns {ExtensiblePromise}
 	 */
-	static all<F extends ExtensiblePromise<T>, T>(iterable: Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F {
+	static all<F extends ExtensiblePromise<T>, T>(iterable: { [_: string]: T | Promise<T> | Thenable<T> } | Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F {
+		if (!isArrayLike(iterable) && !isIterable(iterable)) {
+			let promiseKeys = Object.keys(iterable);
```
agubler: could be a `const`, sorry 🤓

src/async/ExtensiblePromise.ts
```diff
@@ -46,12 +46,33 @@ export default class ExtensiblePromise<T> {
...
-	 * @param iterable    An iterable of values to resolve. These can be Promises, ExtensiblePromises, or other objects
+	 * @example
+	 * ExtensiblePromise.all({ one: 1, two: 2 }).then(results => console.log(results));
+	 * // { one: 1, two: 2 }
+	 *
+	 * @param iterable    An iterable of values to resolve, or a key/value pair of values to resolve. These can be Promises, ExtensiblePromises, or other objects
 	 * @returns {ExtensiblePromise}
 	 */
-	static all<F extends ExtensiblePromise<T>, T>(iterable: Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F {
+	static all<F extends ExtensiblePromise<T>, T>(iterable: { [_: string]: T | Promise<T> | Thenable<T> } | Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F {
```
agubler: Might be nice if we create an interface for the `iterable` type and break out `{  [index: string]: T }` so that it can be used as the `returnValue` on resolution in the `Promise.all(....`

agubler (commented):
Just a couple of comments, there may be other `let`s that can be `const`s too.

src/async/ExtensiblePromise.ts
rorticus: They are different types? The parameter is taking `{ [_: string]: T | Promise<T> | Thenable<T> }` and the return value is just `{ [_: string]: T }`.

One thing we can do (although it has nothing to do with this change) is break out the definition,

```ts
static all<F extends ExtensiblePromise<{ [_: string]: T }>, T>(iterable: { [_: string]: T | Promise<T> | Thenable<T> }): F;
static all<F extends ExtensiblePromise<T[]>, T>(iterable: Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F;
static all<F extends ExtensiblePromise<T>, T>(iterable: { [_: string]: T | Promise<T> | Thenable<T> } | Iterable<(T | Thenable<T>)> | (T | Thenable<T>)[]): F {
```

That would at least give a little bit more definition to the return type?

agubler: That would be nice.

rorticus: OK, added some method overloading :)

agubler (approved):

PR #242, opened 2016-11-17
Removing stringExtras padding methods because you can get them from d…

rorticus:
…ojo-shim/string now, shim issue https://github.com/dojo/shim/issues/27

Note that the removed functionality is now provided by [this PR](https://github.com/dojo/shim/pull/54). So.. this should probably wait until the other PR merged first.


agubler (approved):
Although this is likely to break anyone who is already using `core` and specifically the padding methods.

However as we are still pre beta and are still very fluid I think this is fine to go in.

Issue #139, opened 2016-04-04
Fetch API


kitsonk:
We should integrate the Fetch API as a provider for `request` and use it when available.

PR #243, opened 2016-11-21
Adding fetch implementation of request api, issue #139

rorticus:
Added `request/fetch` for performing requests based on the [`fetch` API](https://fetch.spec.whatwg.org/).

### DIFF
--- a/src/request/fetch.ts
+++ b/src/request/fetch.ts
@@ -104,7 +104,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 
 			const responseHeaders: ResponseHeaders = {};
 
-			forOf(fetchResponse.headers.enrties(), ([key, value]) => {
+			forOf(fetchResponse.headers.entries(), ([key, value]) => {
 				responseHeaders[ key.toLowerCase() ] = value;
 			});
 
--- a/src/request/fetch.ts
+++ b/src/request/fetch.ts
@@ -104,8 +104,8 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 
 			const responseHeaders: ResponseHeaders = {};
 
-			forOf(fetchResponse.headers.keys(), (key: string) => {
-				responseHeaders[ key.toLowerCase() ] = fetchResponse.headers.get(key);
+			forOf(fetchResponse.headers.enrties(), ([key, value]) => {
+				responseHeaders[ key.toLowerCase() ] = value;
 			});
 
 			body.then((body: any) => {
--- a/src/request/fetch.ts
+++ b/src/request/fetch.ts
@@ -75,7 +75,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 			timeout && timeout.destroy();
 
 			let { responseType = '' } = options;
-			let body: any = null;
+			let body: Promise<any> = null;
 
 			switch (responseType) {
 				case 'arraybuffer':
@@ -104,7 +104,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 				responseHeaders[ key.toLowerCase() ] = fetchResponse.headers.get(key);
 			});
 
-			Promise.resolve(body).then((body: any) => {
+			body.then((body: any) => {
 				resolve({
 					statusCode: fetchResponse.status,
 					statusText: fetchResponse.statusText,
--- a/src/request/fetch.ts
+++ b/src/request/fetch.ts
@@ -20,13 +20,17 @@ declare function fetch(_: any): any;
 export interface FetchRequestOptions extends RequestOptions {
 }
 
+interface ResponseHeaders {
+	[name: string]: string;
+}
+
 export default function fetchRequest<T>(url: string, options: FetchRequestOptions = {}): ResponsePromise<T> {
 	const fetchRequestOptions: any = {};
 	const fetchRequestHeaders: Headers = new Headers();
 	const requestUrl = generateRequestUrl(url, options);
 
 	if ((!options.user || !options.password) && options.auth) {
-		let auth = options.auth.split(':');
+		const auth = options.auth.split(':');
 		options.user = decodeURIComponent(auth[ 0 ]);
 		options.password = decodeURIComponent(auth[ 1 ]);
 	}
@@ -50,7 +54,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 		let hasContentTypeHeader = false;
 		let hasRequestedWithHeader = false;
 
-		for (let header in headers) {
+		for (const header in headers) {
 			if (header.toLowerCase() === 'content-type') {
 				hasContentTypeHeader = true;
 			} else if (header.toLowerCase() === 'x-requested-with') {
@@ -74,7 +78,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 		fetch(request).then((fetchResponse: any) => {
 			timeout && timeout.destroy();
 
-			let { responseType = '' } = options;
+			const { responseType = '' } = options;
 			let body: Promise<any>;
 
 			switch (responseType) {
@@ -88,7 +92,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 
 				case 'xml':
 					body = fetchResponse.text().then((asText: string) => {
-						let parser = new DOMParser();
+						const parser = new DOMParser();
 						return parser.parseFromString(asText, 'text/xml');
 					});
 					break;
@@ -98,7 +102,7 @@ export default function fetchRequest<T>(url: string, options: FetchRequestOption
 					break;
 			}
 
-			let responseHeaders: any = {};
+			const responseHeaders: ResponseHeaders = {};
 
 			forOf(fetchResponse.headers.keys(), (key: string) => {
 				responseHeaders[ key.toLowerCase() ] = fetchResponse.headers.get(key);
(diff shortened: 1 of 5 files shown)


src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+	let request = new Request(requestUrl, fetchRequestOptions);
+
+	return new Task<RequestResponse<T>>((resolve, reject) => {
+		let timeout: Handle;
+
+		fetch(request).then((fetchResponse: any) => {
+			timeout && timeout.destroy();
+
+			let { responseType = '' } = options;
+			let body: any = null;
```
agubler: can you type this as a `Promise<any>` (or not `any` if you know the body type) and then you wouldn't need the `Promise.resolve('body').then...` just simply `body.then...`

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+						return parser.parseFromString(asText, 'text/xml');
+					});
+					break;
+
+				default:
+					body = fetchResponse.text();
+					break;
+			}
+
+			let responseHeaders: any = {};
```
agubler: any reason not `const`?

src/request/fetch.ts
rorticus: Yes! Good point. I guess I had typed that before I realized it was going to have to be a Promise, and then forgot to change it :/

src/request/fetch.ts
agubler: Also do you know the type of `headers`, presumably something like maybe

```ts
interface Headers {
    [index string]: string;
}
```

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+					body = fetchResponse.arrayBuffer();
+					break;
+
+				case 'blob':
+					body = fetchResponse.blob();
+					break;
+
+				case 'xml':
+					body = fetchResponse.text().then((asText: string) => {
+						let parser = new DOMParser();
```
agubler: `const`?

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+
+	let request = new Request(requestUrl, fetchRequestOptions);
+
+	return new Task<RequestResponse<T>>((resolve, reject) => {
+		let timeout: Handle;
+
+		fetch(request).then((fetchResponse: any) => {
+			timeout && timeout.destroy();
+
+			let { responseType = '' } = options;
```
agubler: `const`?

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+export interface FetchRequestOptions extends RequestOptions {
+}
+
+export default function fetchRequest<T>(url: string, options: FetchRequestOptions = {}): ResponsePromise<T> {
+	const fetchRequestOptions: any = {};
+	const fetchRequestHeaders: Headers = new Headers();
+	const requestUrl = generateRequestUrl(url, options);
+
+	if ((!options.user || !options.password) && options.auth) {
+		let auth = options.auth.split(':');
```
agubler: `const`?

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	fetchRequestOptions.method = options.method;
+
+	if (options.headers) {
+		const headers = options.headers;
+		let hasContentTypeHeader = false;
+		let hasRequestedWithHeader = false;
+
+		for (let header in headers) {
```
agubler: sorry about all the `const` suggestions...

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+					break;
+			}
+
+			let responseHeaders: any = {};
+
+			forOf(fetchResponse.headers.keys(), (key: string) => {
+				responseHeaders[ key.toLowerCase() ] = fetchResponse.headers.get(key);
+			});
+
+			Promise.resolve(body).then((body: any) => {
```
agubler: call this `data` and you can just use `data` in the resolved object and drop the `: body`

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+						return responseHeaders[ name.toLowerCase() ];
+					}
+				});
+			}, reject);
+		}, reject);
+
+		if (options.timeout > 0 && options.timeout !== Infinity) {
+			timeout = (function (): Handle {
+				const timer = setTimeout(function (): void {
+					const error = new RequestTimeoutError('Request timed out after ' + options.timeout + 'ms');
```
agubler: how about using template literals 

```ts
new RequestTimeoutError(`Request timed out after ${options.timeout}ms`);
```

src/request/fetch.ts
```diff
@@ -0,0 +1,135 @@
...
+					reject(error);
+				}, options.timeout);
+
+				return createHandle(function (): void {
```
agubler: can do if you so desire,

```ts
				return createHandle((): void => {
					clearTimeout(timer);
				});
```

or perhaps even

```ts
			timeout = ((): Handle => {
				const timer = setTimeout((): void => {
					const error = new RequestTimeoutError('Request timed out after ' + options.timeout + 'ms');
					reject(error);
				}, options.timeout);

				return createHandle(clearTimeout.bind(null, timer));
			})();
```

agubler (changes requested):
Just a couple minor comments/questions.

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
rorticus: Done!

src/request/fetch.ts
```diff
@@ -0,0 +1,139 @@
...
+					break;
+
+				default:
+					body = fetchResponse.text();
+					break;
+			}
+
+			const responseHeaders: ResponseHeaders = {};
+
+			forOf(fetchResponse.headers.keys(), (key: string) => {
```
agubler: another mainly styling suggestion (I think that works 😉 )

```ts
			forOf(fetchResponse.headers.entries(), ([key, value]) => {
				responseHeaders[ key.toLowerCase() ] = value;
			});
```

src/request/fetch.ts
```diff
@@ -0,0 +1,139 @@
...
+					break;
+
+				default:
+					body = fetchResponse.text();
+					break;
+			}
+
+			const responseHeaders: ResponseHeaders = {};
+
+			forOf(fetchResponse.headers.enrties(), ([key, value]) => {
```
agubler: oops typo.

rorticus: Ugh. how did that build? ....

agubler: I really do not know!

src/request/fetch.ts
```diff
@@ -0,0 +1,128 @@
...
+				case 'arraybuffer':
+					body = fetchResponse.arrayBuffer();
+					break;
+
+				case 'blob':
+					body = fetchResponse.blob();
+					break;
+
+				case 'xml':
+					body = fetchResponse.text().then((asText: string) => {
```
agubler: I believe `.then` should be on a new line as per the style guide

rorticus: +1

agubler (approved):
Approved from me! Someone else might want eyes on before the big green light.

agubler:
Not sure if anyone else will get the opportunity to review this 🎱

dylans:
A few stray thoughts that are probably worth addressing mostly later as we decide how useful the fetch API might be:

* the fetch API doesn't seem to have a convenience method for json, so the user must explicitly parse json? Is that how our request implementation works also?
* the fetch API seems to implement ReadableStream... I assume we're planning to ignore that for now given the size of the streams shim?
* are we ignoring the type Document (I see we have arraybuffer and blob), or does that get parsed the same way as plain text?
* the fetch API has a number of scenarios around cors, requestType, requestDestination, etc. Do we plan to/need to support these features now or in the future?

rorticus:
> the fetch API doesn't seem to have a convenience method for json, so the user must explicitly parse json? Is that how our request implementation works also?

We have [a response filter](https://github.com/dojo/core/blob/master/src/request.ts#L198-L207) that runs at a "higher level" that will turn the json type into an actual JSON object.

> the fetch API seems to implement ReadableStream... I assume we're planning to ignore that for now given the size of the streams shim?

I'm OK with ignoring that :)  If we do decide to implement it, we should at least figure out how we're going to pull in `streams` without the dependency in the node request.

> are we ignoring the type Document (I see we have arraybuffer and blob), or does that get parsed the same way as plain text?

It'll get parsed as plain text.

> the fetch API has a number of scenarios around cors, requestType, requestDestination, etc. Do we plan to/need to support these features now or in the future?

That is a great question :) I've ignored them for now because we didn't have anything similar in the xhr/node requests, but adding those options might be a compelling reason why someone would _want_ to use the fetch API over the alternatives.

rorticus:
Dylan's comments have been split into https://github.com/dojo/core/issues/255 and https://github.com/dojo/core/issues/254 to get this landed.

PR #244, opened 2016-11-21
Adding v4 UUID generation, issue 214

rorticus:
I decided to not go with using `node-uuid`.. My thinking was that this is a pretty small bit of code that will probably never change, so it seems like a lot less overhead to just add it here and forget about it.

I made this a separate module (instead of putting it in `utils`) because I thought perhaps in the future we would want to offer more robust UUID generation (which could then maybe just use `node-uuid`).

### DIFF
--- a/src/uuid.ts
+++ b/src/uuid.ts
@@ -1,0 +1,13 @@
+// implementation of v4 UUID from https://github.com/kitsonk/hokan/blob/master/util/main.js
+
+/**
+ * Returns a v4 compliant UUID.
+ *
+ * @returns {string}
+ */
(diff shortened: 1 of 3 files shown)


src/uuid.ts
```diff
@@ -0,0 +1,13 @@
+// implementation of v4 UUID from https://github.com/kitsonk/hokan/blob/master/util/main.js
```
agubler: lets lose this comment.

rorticus: lost :)

agubler (approved):

PR #247, opened 2016-11-23
Adding readonly modifier where appropriate, issue #180

rorticus:
Added `readonly` modifier where I felt it was appropriate. (issue #180)

At first I was adding it to all interfaces where the declared properties where not being written to, but I decided not to do that, so people could still modify objects that they created.

```ts
let options: KwArgs {
	year: 2016,
	month: 10
};

if(useLastMonth) {
	options.month = 9;
}

const date = new DateObject(options);
```

instead of,

```ts
let month = 10;
if(useLastMonth) { 
	month = 9;
}

let options: KwArgs {
	year: 2016,
	month
};

const date = new DateObject(options);
```

Granted, this is a terrible, contrived, example, but you get the idea.

### DIFF
--- a/src/DateObject.ts
+++ b/src/DateObject.ts
@@ -25,10 +25,10 @@ export interface OperationKwArgs {
  */
 export interface DateProperties {
 	dayOfMonth: number;
-	dayOfWeek: number;
-	daysInMonth: number;
+	readonly dayOfWeek: number;
+	readonly daysInMonth: number;
 	hours: number;
-	isLeapYear: boolean;
+	readonly isLeapYear: boolean;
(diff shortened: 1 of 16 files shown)


agubler:
@rorticus how do we know that this won't break consumers already using this package and updating value we've now declared `readonly`?

rorticus:
@agubler , I've tried to leave "public interfaces", where people pass in configuration objects, alone. If you see a place where I missed that constraint, let me know. Most of these changes should just be for internally used interfaces and private variables that hopefully people aren't messing with anyways.

agubler (approved):

Issue #201, opened 2016-10-04
dojo-core/load and non-AMD environments


kitsonk:
We likely need to enhance `dojo-core/load` so that it can load resources via other loaders (e.g. in a WebPack bundle) and have that be transparent to the end developer.

We also need to consider how `load()` would handle something that contains an AMD plugin signature, in that if it is not running under `AMD` should it try to replicate the functionality of the AMD plugin interface, loading the plugin module and executing the plugin API as specified in the AMD specification?  This feature would make it easier for end developers to potentially use AMD plugins without having to worry about where their code ends up...

mwistrand:
Regarding an AMD-style plugin implementation, I do not think the [amdjs API](https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md) could be met exactly given that `@dojo/core/load` is not actually an AMD loader, but we should be able to get close:

```typescript
interface LoadPlugin<T> {
  normalize?: (resourceId: string, normalize: (resourceId: string) => string): string;
  load(resourceId: string, load: Load, resolve: (value?: T | Thenable<T>) => void): T;
}
```

Either the default export (`exports.default` in the ES5-transpiled TS) could be a `LoadPlugin` object, or the result of calling `load` would need to a `LoadPlugin` object. The optional `normalize` method would be passed the resource ID, as well as the `require.toUrl` or `require.resolve` method, depending on the environment. The required `load` method would not receive the contextual require, but would instead receive the `@dojo/core/load` method itself, along with a callback that should receive the final value.

Since there is now a [baked-in webpack-specific `load`](https://github.com/dojo/core/blob/master/src/load/webpack.ts) method, any updates to that module would be with the caveat that any value that could potentially be loaded via a plugin would need to be already included in a webpack chunk.

dylans:
> Regarding an AMD-style plugin implementation, I do not think the amdjs API could not be met exactly given that @dojo/core/load is not actually an AMD loader

Did you mean to have a double negative here? :)

mwistrand:
@dylans No, thanks for catching that. I meant to say that I don't believe that API can be fulfilled exactly as specified.

mwistrand:
[...]

PR #249, opened 2016-12-05
Add webpack-specific `load`.

mwistrand:
Addresses a portion of #201. Given that webpack 1) does not understand contextual `require` passed as a value, and 2) needs to know about any modules that _could_ be required, and therefore must include them in one of its known chunks, the current `dojo-core/load` implementation is incompatible with webpack.

To get around this, a replacement module can be used that maps resolved module paths to webpack's numerical module IDs. This new `load-webpack` module will not be used directly within applications, but instead will be used by a webpack plugin that generates its module information and uses it as a drop-in replacement for `dojo-core/load`. It is included in `dojo-core` both to keep it side-by-side with and ensure it is transpiled consistently with the main `load` implementation.


tests/unit/load-webpack.ts
```diff
@@ -0,0 +1,145 @@
...
+		},
+
+		'relative path (up two directories)'() {
+			return load(() => '/path/to/nested/module', '../../bar').then((results: any[]) => {
+				assert.strictEqual(results.length, 1);
+				assert.strictEqual(results[0].bar, 'baz');
+			});
+		},
+
+		'relative path (beyond root directory)'() {
```
matt-gadd: 👍 nice test, i hadn't even thought about that when i'd poc'd it. do you think we should go with the behaviour of using the top directory, or do you think we should align with the dojo loader (at least the old one I think?) which threw an irrational path error?

tests/unit/load-webpack.ts
```diff
@@ -0,0 +1,145 @@
...
+let webpackModules: WebpackModules;
+function setModules(modules?: { [mid: string]: any }) {
+	let nextId = 0;
+	webpackModules = Object.create(null) as WebpackModules;
+
+	if (modules) {
+		const idMap = Object.create(null);
+		Object.keys(modules).forEach((mid: string) => {
+			const id = nextId++;
+			idMap[mid] = { id, lazy: mid.indexOf('bundle!') > -1 };
```
matt-gadd: for the contents of the map, do you think we might want to add some mechanism for different behaviours per plugin perhaps? rather than just the lazy flag at the moment that we have for `bundle`?

tests/unit/load-webpack.ts
mwistrand: Hmm. That's a better solution, actually, then trying to "correct" the path. I'll update the code/tests.

tests/unit/load-webpack.ts
mwistrand: One option would be to inject a `__loaders__` object that gets passed the `require` function, `mid`, and `resolve` callback, with `load` reducing the value for multiple loaders as needed. We could provide `bundle` out of the box, but others would need to be specified via a config. Are there any security considerations that would need to be addressed with such an implementation?

tests/unit/load-webpack.ts
mwistrand: Code/tests updated to throw an error with a relative path that traverses beyond the root directory.

tests/unit/load-webpack.ts
mwistrand: Also, is this an issue that should be resolved before the PR lands, or should/can it be postponed for an ongoing discussion?

matt-gadd: sorry for the delay in replying on this. given its seemingly an implementation detail (of some kind of internal registry). I think it would be a nice to have in the future - that wouldn't block this PR.

matt-gadd:
do you think we should move `load-webpack` to a folder? ie `load/webpack.ts`, that way we can avoid the dash?

mwistrand:
Yes, that would be better, especially when considering that it is already what we do for the request modules.

matt-gadd (approved):

Issue #59, opened 2015-07-27
AMD Plugin Interface


kitsonk:
Shouldn't we put the AMD plugin interface as a specified interface either in core, or dojo/loader?

I noticed that in the core/has implementation, it isn't relying upon an interface, therefore potentially exposing an API that isn't consistent.  One of the strengths of TypeScript is being able to collate these interfaces and expose them to others.

kfranqueiro:
I would think definitions for this would belong in dojo/loader, except if we still have any thought of being able to interoperate with other loaders, in which case adding an unconditional dependency on a loader you might not use would seem undesirable, and core would be the next best place since most other packages are likely to depend on that.

@edhager, @bryanforbes, thoughts?

kfranqueiro:
Also, would this idea actually _work_ for modules, especially in cases where we're exporting individual functions?  (i.e. is it even feasible to say that a module's exports collectively implement an interface?)

kitsonk:
Well it appears it exists partially in [core/load](https://github.com/dojo/core/blob/master/src/load.ts#L8).  I suspect we just need to move some/all of these interfaces somewhere else and `dojo/has` should implement them.

kitsonk:
`load` is missing a Normalize interface though.

kfranqueiro:
Isn't that just an interface for AMD's `require` as called by consumers, and not specifically for plugins?

kitsonk:
DOH!  You are correct, though I did find it in [dojo/loader](https://github.com/dojo/loader/blob/b9ccd4974258035a52e998a093825d3863de245b/src/loader.ts#L30) though.  It seems the whole thing is already there, but I know we were trying to avoid a hard dependency on the loader.  Is it better located somewhere else?

bryanforbes:
I wonder if, like `lib.d.ts` or `node.d.ts`, there needs to be an `amd.d.ts` that can be shared by projects that will work with any AMD loader.

msssk:
Currently, `dojo-core` has an incomplete `AMDRequire` interface defined in `load.ts`. There is a more complete interface defined in `dojo/loader`, so should we use that ([PR](https://github.com/dojo/core/pull/58))? Or should we extricate AMD definitions into `amd.d.ts`? Where would `amd.d.t.s` live (what project)?

kitsonk:
We can close this issue for now and revisit if needed.

PR #250, opened 2016-12-07
Adding extended observable to implement some RxJs methods, shim #59

rorticus:
Adding a more usable version of the `Observable` we added in dojo/shim.  Imports and naming conventions should be similar to RxJs, so most usages can just be replaced with the new import `dojo-core/Observable`.

Added `toPromise`, `defer`, `map`, `filter`, `toArray`, and `mergeAll`.

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -1,0 +1,179 @@
+import ObservableShim, { ObservableObject, Subscribable, SubscriptionObserver } from 'dojo-shim/Observable';
+import Promise from 'dojo-shim/Promise';
+import { Iterable } from 'dojo-shim/iterator';
+
+function isSubscribable(object: any): object is Subscribable<any> {
+	return object && object.subscribe !== undefined;
+}
+
+export default class Observable<T> extends ObservableShim<T> {
+
+	static of<T>(...items: T[]): Observable<T> {
+		return <Observable<T>> super.of(...items);
+	}
+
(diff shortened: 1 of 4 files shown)


rorticus:
Note that the test failures here are from the fact that `Observables` in dojo/shim aren't released yet.

src/Observable.ts
```diff
@@ -0,0 +1,179 @@
...
+
+		if (typeof mapFunction !== 'function') {
+			throw new TypeError('Map parameter must be a function');
+		}
+
+		return new Observable<U>((observer: SubscriptionObserver<U>) => {
+			self.subscribe({
+				next(value: T) {
+					try {
+						let result: U = mapFunction(value);
```
agubler: `const`

src/Observable.ts
```diff
@@ -0,0 +1,179 @@
...
+				}
+			});
+		});
+	}
+
+	toArray(): Observable<T[]> {
+		const self = this;
+
+		return new Observable<T[]>(observer => {
+			let values: T[] = [];
```
agubler: `const`

agubler (commented):
nice! that standard couple of `let`s that can be `const`s and we can get this landed.

src/Observable.ts
rorticus: Dang. I was so careful this time!

src/Observable.ts
rorticus: ^^ fixed

matt-gadd (approved):

PR #257, opened 2016-12-22
Type updates for ExtensiblePromise and Task

rorticus:
There were some issues with `ExtensiblePromise` and `Task` not returning the right types from `then`, `resolve`, and `all`.  

This hopefully fixes those problems :)

### DIFF
--- a/src/async/ExtensiblePromise.ts
+++ b/src/async/ExtensiblePromise.ts
@@ -139,6 +139,7 @@ export default class ExtensiblePromise<T> {
 	 *
 	 * @returns {ExtensiblePromise}
 	 */
+	then<U, V>(onFulfilled: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected: (reason: Error) => (V | Thenable<V>)): ExtensiblePromise<U | V>;
 	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => void): ExtensiblePromise<U>;
 	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => (U | Thenable<U>)): ExtensiblePromise<U> {
 		const e: Executor<U> = (resolve, reject) => {
--- a/src/async/ExtensiblePromise.ts
+++ b/src/async/ExtensiblePromise.ts
@@ -126,8 +126,8 @@ export default class ExtensiblePromise<T> {
 	 *
 	 * @returns {ExtensiblePromise}
 	 */
-	catch<U>(onRejected: (reason: Error) => (U | Thenable<U>)): ExtensiblePromise<U>;
-	catch<U>(onRejected: (reason: Error) => void): ExtensiblePromise<U> {
+	catch(onRejected: (reason: Error) => T | Thenable<T> | void): ExtensiblePromise<T>;
+	catch<U>(onRejected: (reason: Error) => U | Thenable<U>): ExtensiblePromise<U> {
 		return this.then<U>(undefined, onRejected);
 	}
 
--- a/src/async/ExtensiblePromise.ts
+++ b/src/async/ExtensiblePromise.ts
@@ -18,7 +18,7 @@ function unwrapPromises(iterable: Iterable<any> | any[]): any[] {
 }
 
 export type DictionaryOfPromises<T> = { [_: string]: T | Promise<T> | Thenable<T> };
-export type ListOfPromises<T> = Iterable<(T | Thenable<T>)> | (T | Thenable<T>);
+export type ListOfPromises<T> = Iterable<(T | Thenable<T>)>;
 
 /**
  * An extensible base to allow Promises to be extended in ES5. This class basically wraps a native Promise object,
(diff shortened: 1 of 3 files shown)


src/async/ExtensiblePromise.ts
```diff
@@ -124,8 +126,8 @@ export default class ExtensiblePromise<T> {
 	 *
 	 * @returns {ExtensiblePromise}
 	 */
-	catch<U>(onRejected: (reason: Error) => (U | Thenable<U>)): this;
-	catch<U>(onRejected: (reason: Error) => void): this {
+	catch<U>(onRejected: (reason: Error) => (U | Thenable<U>)): ExtensiblePromise<U>;
+	catch<U>(onRejected: (reason: Error) => void): ExtensiblePromise<U> {
```
nicknisi: I think there is a problem with the types here still. Consider the following example:

```typescript
import ExtensiblePromise from 'dojo-core/async/ExtensiblePromise';

export function booleanPromise(): ExtensiblePromise<boolean> {
	const promise = new ExtensiblePromise(resolve => {
		resolve(true);
	});

	return promise.catch(error => { console.log(error); });
}
```

In this example, the `catch` call is changing the promise type from `ExtensiblePromise<boolean>` to `ExtensiblePromise<void>`, which should not be correct.

I think the following method type definition should work:

```typescript
catch(onRejected: (reason: Error) => T | Thenable<T> | void): ExtensiblePromise<T>;
catch<U>(onRejected: (reason: Error) => U | Thenable<U>): ExtensiblePromise<U> {
```

src/async/ExtensiblePromise.ts
```diff
@@ -137,8 +139,8 @@ export default class ExtensiblePromise<T> {
 	 *
 	 * @returns {ExtensiblePromise}
 	 */
-	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => void): this;
-	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => (U | Thenable<U>)): this {
+	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => void): ExtensiblePromise<U>;
+	then<U>(onFulfilled?: ((value: T) => (U | Thenable<U> | undefined)) | undefined, onRejected?: (reason: Error) => (U | Thenable<U>)): ExtensiblePromise<U> {
```
nicknisi, line 144: I think there might be a problem with the types here as well. Consider the following example:

```typescript
import ExtensiblePromise from 'dojo-core/async/ExtensiblePromise';

const promise = new ExtensiblePromise((resolve, reject) => {
	reject(true);
});

promise.then(() => {
	// return nothing
}, () => 'some string') // catch returns a string
.then(value => {
	console.log('length', value.length);
	// Property 'length' doesn't exist on type 'never';
});
```

The `then` call always assumes the new Promise will have the same type as the `onFulfilled` callback, and never `onRejected`. Comparing this to how ES2015 Promises work, I wouldn't expect the `Property 'length' doesn't exist on type 'never';` to be there, as it's not guaranteed that the type is `never`. I haven't thoroughly tested this, but comparing our definition to the ES2015 Promise definition, this might be a solution:

```typescript
then(onFulfilled?: ((value: T) => T | Thenable<T>) | void, onRejected?: ((reason: Error) => T | Thenable<T>) | void): ExtensiblePromise<T>;
then<U>(onFulfilled: ((value: T) => Thenable<T>) | void, onRejected: (reason: Error) => U | Thenable<U>): ExtensiblePromise<T | U>;
then<U>(onFulfilled: (value: T) => U | Thenable<U>, onRejected?: ((reason: Error) => U | Thenable<U>) | void): ExtensiblePromise<U>;
then<U, V>(onFulfilled: (value: T) => U | Thenable<U>, onRejected: (reason: Error) => V | Thenable<V>): ExtensiblePromise<U | V> {
	const e: Executor<U | V> = (resolve, reject) => {
		function handler(rejected: boolean, valueOrError: T | U | Error) {
			const callback: ((value: T | U | V | Error) => (U | Thenable<U> | V | Thenable<V> | void)) | undefined = rejected ? onRejected : onFulfilled;

			if (typeof callback === 'function') {
				try {
					resolve(<U> callback(<T> valueOrError));
				}
				catch (error) {
					reject(error);
				}
			}
			else if (rejected) {
				reject(valueOrError);
			}
			else {
				resolve(<U> valueOrError);
			}
		}

		this._promise.then(handler.bind(null, false), handler.bind(null, true));
	};

	return new (<{ new(executor: Executor<U | V>): any }> this.constructor)(e);
}
```

rorticus: Good catch!  I'll take a look on Monday!

rorticus: OK @nicknisi , go ahead and take a look now. I used something similar to your approach and got your example working.

nicknisi:
👍 This is looking good!

nicknisi:
It looks like the `catch` signatures may need to be updated for Task as well.

```typescript
import Task from 'dojo-core/async/Task';

// This one throws an error
// Type 'Task<void>' is not assignable to type 'Task<string>'. Type 'void' is not assignable to type 'string'.
export function returnString(): Task<string> {
	return Task.resolve(undefined)
		.then(() => 'hi')
		.catch(error => {
			// do nothing
		});
}

// no errors because it throws
export function returnString2(): Task<string> {
	return Task.resolve(undefined)
		.then(() => 'hi')
		.catch(error => {
			throw error;
		});
}

// no errors because it returns a string
export function returnString3(): Task<string> {
	return Task.resolve(undefined)
		.then(() => 'hi')
		.catch(error => {
			return 'bye';
		});
}
```

rorticus:
@nicknisi I updated the `catch` types for `Task` as well.

nicknisi:
Types are working better now! Thank you!

nicknisi (approved):

Issue #253, opened 2016-12-13
Simplify core/request


rorticus:
Currently the `request` method will dynamically load the default provider, depending on if you are running in node or the browser. This adds some complexity to the code and probably isn’t even needed.

We’re looking for a usage like this,

```js
import request from 'dojo-core/request';
// import node support manually if we want to use it
import nodeProvider from 'dojo-core/request/node';

// set the default provider some how
request.setDefaultProvider(nodeProvider);

// the next request will go to node
request.get(...);
```

* Remove the dynamic loading from `ProviderRegistry` and just assume that `defaultValue` is already a provider, not a string.
* Add a method to set the default provider, so it can be injected.
* Create a convenience method to inject `request/node` as the default provider for people running in Node.

You can see a rough implementation (as well as some unrelated changes...) here,

https://github.com/dojo/core/compare/master...rorticus:no-streams

sebilasse:
Just my 2 cents : I like the dynamic loading.
It is consistent with all modules which you can use either in the browser or node.js ...
e.g. /text, /util, /queue, /on have switches for node/browser and as a user
I would expect to use any module without external requirements.

matt-gadd:
@sebilasse We would still support dynamic loading (even when in most environments it is definitely more desirable to make concrete at build time), it just wouldn't happen in the core of `request` itself. Currently the code to dynamically require down the browser path is super ugly - we queue requests while loading the provider, this kind of thing doesn't seem like it should be a `request` concern.

matt-gadd:
Following on: In the past, the above complications would've been removed by dealing with this in the dependency block in `AMD` using the `has` plugin to conditionally require the environment specific module. But given we're consciously trying to keep as format neutral for the future we're better placed on making these kind of module loading decisions outside of `request`. The point being that the environment will/should be known prior to a request being made.

rorticus:
Fixed in #261

PR #261, opened 2017-01-06
Updating request library to mimic fetch API

rorticus:
This PR updates the request APIs to mimic the fetch APIs, which makes it possible to do cool stuff like look at the response headers _before_ the body is downloaded.

This also introduces an event based `Response` object that can be used to monitor the download progress and send responses to streams (a helper utility for that will be provided in `dojo-streams`).

Issue: https://github.com/dojo/core/issues/253


src/request.ts
```diff
@@ -1,207 +1,47 @@
...
-				return inTest.test(url);
-			};
-		}
-		else {
-			entryTest = <RequestFilterTest> inTest;
-		}
+export const providerRegistry = new ProviderRegistry();
 
-		return super.register(entryTest, value, first);
+const request: {
```
rishson, line 8: Any reason why we don't also support `OPTIONS` and `HEAD`? At the moment we support a list that is neither the CORS safelist of verbs or the verbs subject to normalization in the fetch spec.

rorticus: Not a good reason that I know of 😄 You can still still make a custom request with whatever method you want, but adding `OPTIONS` and `HEAD` should be fairly trivial, so we can do that too!

rishson: Yeah, given we actually normalize, e.g. send `POST` for a `post()`, then probably best we support all the verbs that spec recommends normalization for. As you say, nothing to stop them doing a custom request for `post` or `pOsT`..

rorticus: Added 👍

rishson: 🔥 
TVM!

dylans: Original reason was likely based on what was done in Dojo 1, fwiw. :)

dylans (approved):
I think I'm ok with this... should have others review as there's a good risk of it causing breaking changes.

agubler:
I don't think we have anything dojo 2 using request, so i'm also okay with the changes. See you've asked for a review from @bryanforbes, that's probably a good idea 👍

src/request/providers/node.ts
```diff
@@ -0,0 +1,558 @@
...
+	}
+}
+
+function redirect(resolve: (p?: any) => void, reject: (_?: Error) => void, originalUrl: string, redirectUrl: string | null, options: NodeRequestOptions): boolean {
+	if (!options.redirectOptions) {
+		options.redirectOptions = {};
+	}
+
+	const { limit: redirectLimit = DEFAULT_REDIRECT_LIMIT } = options.redirectOptions;
+	const { count: redirectCount = 0 } = options.redirectOptions;
```
bryanforbes: Would it be clearer and/or more efficient to format these two lines as such?

```ts
const {
    limit: redirectLimit = DEFAULT_REDIRECT_LIMIT,
    count: redirectCount = 0
} = options.redirectOptions;
```

src/request/node.ts
```diff
@@ -1,334 +0,0 @@
-import Task from '../async/Task';
```
bryanforbes, line 1: In order for git to recognize the move of this file, it should be moved in one commit and changed in another. I would recommend rewriting history in this branch to add a commit before starting that moves files into place before making changes.

src/request/providers/node.ts
```diff
@@ -0,0 +1,558 @@
...
+			const response = new NodeResponse(message);
+
+			// Redirection handling defaults to true in order to harmonise with the XHR provider, which will always
+			// follow redirects
+			if (
+				response.status >= 300 &&
+				response.status < 400
+			) {
+				const redirectOptions = options.redirectOptions || {};
+				const newOptions = Object.create(options);
```
bryanforbes: This creates a shallow copy of `options`, so assigning to a sub-object (as is being done on line 227 above) could potentially modify the original. Since no such function exists in `@dojo/core` to do a deep create, we should create a PR for one and use it here.

src/request/providers/node.ts
```diff
@@ -0,0 +1,558 @@
...
+	}).catch(function (error: Error): any {
+		const parsedUrl = urlUtil.parse(url);
+
+		if (parsedUrl.auth) {
+			parsedUrl.auth = '(redacted)';
+		}
+
+		const sanitizedUrl = urlUtil.format(parsedUrl);
+
+		error.message = '[' + requestOptions.method + ' ' + sanitizedUrl + '] ' + error.message;
```
bryanforbes, line 567: Are there cases where setting `error.message` would throw?

src/request/providers/xhr.ts
```diff
@@ -0,0 +1,310 @@
...
+
+			return data;
+		});
+	};
+}
+
+function noop () {}
+
+function setOnError(request: XMLHttpRequest, reject: Function) {
+	request.onerror = function (event) {
```
bryanforbes: Should this use `addEventListener()`?

src/request/providers/xhr.ts
```diff
@@ -0,0 +1,310 @@
...
+	const request = new XMLHttpRequest();
+	const requestUrl = generateRequestUrl(url, options);
+
+	options = Object.create(options);
+
+	if (!options.method) {
+		options.method = 'GET';
+	}
+
+	if ((!options.user || !options.password) && options.auth) {
```
bryanforbes: What is the reasoning behind allowing two ways to pass authentication information?

src/request/providers/xhr.ts
```diff
@@ -0,0 +1,310 @@
...
+				return;
+			}
+
+			if (request.readyState === 2) {
+				const response = new XhrResponse(request);
+
+				const task = new Task<XMLHttpRequest>((resolve, reject) => {
+					timeoutReject = reject;
+
+					request.onprogress = function (event: any) {
```
bryanforbes, line 195: This only handles download progress. It would be nice (perhaps in another PR) to handle upload progress as well.

src/request/providers/xhr.ts
```diff
@@ -0,0 +1,310 @@
...
+		};
+
+		setOnError(request, reject);
+
+	}, abort);
+
+	request.open(options.method, requestUrl, !options.blockMainThread, options.user, options.password);
+
+	if (has('filereader') && has('blob')) {
+		request.responseType = 'blob';
```
bryanforbes, line 257: Does this need to be set before calling `request.open()`?

src/QueuingEvented.ts
```diff
@@ -0,0 +1,67 @@
+import Evented from './Evented';
```
bryanforbes, line 2: Please add some documentation to this class about what it does.

src/request/providers/node.ts
agubler: would we not just use `deepAssign({}, options)`?

rorticus: I'll go with that @agubler

src/request/providers/node.ts
rorticus: Maybe, but either way we're throwing an error. I could swallow errors to `error.message` and just ignore it if it didn't work?  What do you think?

src/request/providers/node.ts
bryanforbes: `deepAssign({}, options)` would also work. I figured there was a reason the code was using `Object.create()` instead of `Object.assign()`, so to keep the code somewhat the same, I was noting that we don't have the `Object.create()` counterpart in `@dojo/lang`. Once you start iterating over properties, you may as well use `deepAssign()` anyway.

src/request/providers/node.ts
bryanforbes: I'm really not sure what the best practice is here. At this point, it might be best to just leave this and if we hit this in reality (rather than trying to cover every potential case), we can adjust.

src/request/providers/node.ts
rorticus: Updated!

src/request/providers/node.ts
rorticus: OK. Let's ignore this for now.

src/QueuingEvented.ts
rorticus: Added!

src/request/node.ts
rorticus: So.. I played around with a few ways to get git to recognize this as a rename rather than a new file. I think that it worked.. but it dawned on me that since we only squash from github, that special 'renaming commit' will get squashed into the other commits and we'd lose that history anyways, ending up with a single "'delete xyz', 'add xyy'" flow with files so different that git won't think they are a rename (the problem we have now).

src/request/providers/xhr.ts
rorticus: https://github.com/dojo/core/issues/285

src/request/providers/xhr.ts
rorticus: This has been addressed!

src/request/node.ts
bryanforbes: Oh right. Then you can safely ignore me.

bryanforbes (approved):
👍

Issue #210, opened 2016-10-11
load() should handle synthetic default imports


kitsonk:
When load() loads a module, is should determine if there is meta data on the module that signals that is an ES Module that has been transpiled and resolve to the default export if this is the case.  The API should also be enhanced to allow that default behaviour to be avoided as well.

Refs dojo/app#74

Issue #202, opened 2016-10-04
dojo-core/load and alignment for ES import()


kitsonk:
There is a Stage 2 proposal for [`import()`](https://github.com/domenic/proposal-import-function) which would essentially be the same as our `load`.

We should potentially consider aligning to that proposal and potentially migrating it to `dojo-shim`.

dylans:
Now at stage 3, we should decide what to do with this.

dylans:
Issue moved to [dojo/meta #133](https://github.com/dojo/meta/issues/133) via [**ZenHub**](https://www.zenhub.com/)

dylans:
Per @matt-gadd : I don't believe we can implement imports here in dojo/core. Mainly due to us needing a contextual require to work with relative paths for AMD. The imports topic is probably a larger discussion in dojo/meta and dojo/loader.

PR #262, opened 2017-01-06
Load enhancements

lzhoucs:
Enhance `load()` with options to load based on `__esModule` metadata either: a) default export. b) all exports except default. The default behavior is a). Existing code spreading across all other packages that depends on `load()` will have to update as well.

Resolves: #210

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -46,17 +46,26 @@ function processModules(modules: any, isDefault: boolean): Object | Object[] {
 }
 
 const load: Load = (function (): Load {
-	if (typeof module === 'object' && typeof module.exports === 'object') {
-		return function (contextualRequire: any, isDefault: any, ...moduleIds: string[]): Promise<any[]> {
-			if (typeof isDefault !== 'boolean') {
+	return function (contextualRequire: any, isDefault: any, ...moduleIds: string[]): Promise<any[]> {
+		if (typeof contextualRequire === 'boolean') {
+			if (typeof isDefault === 'string') {
 				moduleIds.unshift(isDefault);
-				isDefault = true;
 			}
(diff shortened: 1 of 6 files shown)


matt-gadd:
I do wonder whether this is the correct approach for the api. `isDefault` (which I think should be named `hasDefault` if staying as is), can only be set for all the modules as `load` can take more than 1 `mid`, which seems inflexible. It seems like `load` should perhaps be left as is today, and only care about actually requiring the module. 

Would it be nicer to have a functional utility(s) that you can use after `load`, to pick the `defaults` out? Something like:

for all modules:
```js
load('a', 'b', 'c')
    .then(useDefault)
    .then(([ a, b, c ]) => {
        // done 
    });
```

for specific modules:
```js
load('a', 'b', 'c')
    .then(([ a, b, c ]) => [ a, useDefault(b), c ])
    .then(([ a, b, c ]) => {
        // done
    });
```

it also means we don't have to do any annoying additional argument shifting/swapping due to the multiple `load` function overloads.

dylans:
While in the context of this topic, it may be worth looking into imports() which is now stage 3, which is tracked in #202?

matt-gadd:
@dylans I don't believe we can implement `imports` here in `dojo/core`. Mainly due to us needing a contextual `require` to work with relative paths for `AMD` so we can't match the signature of `imports`. The `imports` topic is probably a larger discussion in `dojo/meta` and `dojo/loader`.

dylans:
@matt-gadd ok, I will move the core issue to meta and add your comment there.

lzhoucs:
Thanks! I do like Matt's idea and Rory's solution better! The solution is simpler and more flexible.

PR #263, opened 2017-01-07
Clean up `has` module.

lzhoucs:
It is good to have all `has` detection defined in one place for internal use. It is a minor refactoring.

### DIFF
--- a/src/has.ts
+++ b/src/has.ts
@@ -23,3 +23,7 @@ add('xhr2-blob', function () {
 	request.abort();
 	return request.responseType === 'blob';
 });
+
+add('node-buffer', 'Buffer' in global && typeof global.Buffer === 'function');
+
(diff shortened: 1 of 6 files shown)


dylans (changes requested):
We should land this... has a conflict and needs a rebase.

lzhoucs:
I did a rebase. The new Codecov is being slightly confusing. I don't see how my change will decrease coverage. :)

Update: it turns out the Codecov data was out dated after I rebased it. The [latest info](https://github.com/dojo/core/pull/263#issuecomment-276572795) is correct.

dylans (approved):

dylans:
@lzhoucs go ahead and land it, thanks!

lzhoucs:
@dylans , I don't think I have permission yet...

PR #264, opened 2017-01-09
Task: prevent finally w/canceled from resolving

nicknisi:
Prevent scenario where finally call after cancel results in a resolved Task.

Currently, `Task#finally` will return a new resolved Task when the task has already been cancelled. This leads to some inconsistencies where the Task returned by Finally will either be resolved or will not be resolvable, all depending on when `cancel` is called. 

```typescript
// This will return a new Task that is resolved to the return 
// value of the callback, allowing for further chaining of then calls
let task = new Task((resolve) => {
	setTimeout(resolve.bind(null, 'test'), 10);
});
task.cancel();
task = task.finally(() => 'changed');
task.then((value: string) => {
       // This callback will be called, resolving to the return value
       // of the finally call
	console.log(value); // changed
});

// However, when cancel is called after the finally callback is setup,
// the subsequent then call is never run.

let task2 = new Task((resolve) => {
	setTimeout(resolve.bind(null, 'test'), 10);
});
task2 = task2.finally(() => 'changed');
task2.then((value: string) => {
       // This is never resolved
	console.log(value);
});
task2.cancel();
```

This fixes the inconsistency by instead calling the callback passed to `finally` and then returning the same Task instance, which won't resolve if the Task has been cancelled.

### DIFF
--- a/src/async/Task.ts
+++ b/src/async/Task.ts
@@ -88,7 +88,7 @@ export default class Task<T> extends ExtensiblePromise<T> {
 	/**
 	 * The finally callback for this Task (if it was created by a call to `finally`).
 	 */
-	private _finally: () => any;
+	private _finally: () => void;
 
 	/**
 	 * The state of the task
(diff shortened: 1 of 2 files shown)


rorticus (approved):
Looks good! 👍

dylans (approved):
Ci failure on Android is unrelated to this code, so let's just land it and either remove Android from the list or find the issue in a separate PR.

Issue #144, opened 2016-04-19
Merge compose/aspect.ts and core/aspect.ts


kitsonk:
We should rationalise [compose/aspect](https://github.com/dojo/compose/blob/master/src/aspect.ts) and core/aspect.

Relates to #61 in what is the API of core/aspect.

kitsonk:
We seem to be surviving on this fine... need to do one day though, but not a beta blocker.

dylans:
For API sanity, it's probably worth cleaning this up for beta?

PR #269, opened 2017-01-18
Merging compose/aspect with core/aspect

rorticus:
Issue #144 

This should be backwards compatible with the original `after`, `before`, and `around` methods provided by `core/aspect`, as well as be able to accept functions with no objects (from `compose/aspect`).


dylans (approved):
Looks fine to me, as this is mostly ported from dojo/compose's advice implementation.

rorticus:
@agubler would you mind giving this a look?

agubler (approved):
no problem from me

PR #270, opened 2017-01-18
Adding loader helper to load ES modules, issue #210

### LINKED ISSUE #210: load() should handle synthetic default imports

rorticus:
Adding a helper to make loading ES modules easier. This is based on @matt-gadd 's suggestion in https://github.com/dojo/core/pull/262

```ts
import load, { useDefault } from '@dojo/core/load';

load('es-module').then(useDefault).then([module] => { /* do something */ });
```

`useDefault` will check if a module has the `__esModule` meta data and if so, return the module's `default` value.

Can also be used on an individual module,

```ts
load('a', 'b').then([a, b] => [useDefault(a), b]).then([moduleA, moduleB] => { /* do something */ });
```

Resolves #210

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -1,6 +1,6 @@
+import { forOf, isIterable, isArrayLike } from '@dojo/shim/iterator';
 import Promise from '@dojo/shim/Promise';
 import { Require } from '@dojo/interfaces/loader';
-import { isArray } from 'util';
 
 declare const require: Require;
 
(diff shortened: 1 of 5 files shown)


dylans (approved):
Only thought I had was, is it worth supporting iterables. If so might as well. :)

rorticus:
Added support for iterables!

dylans (approved):

matt-gadd (approved):

PR #271, opened 2017-01-19
Implement AMD-style plugins for `load` and `load/webpack`.

mwistrand:
Updates `load.ts` and `load/webpack.ts` to allow [amdjs-style plugins](https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md) to be used.

Resolves a portion of #201.

```ts
// html-dom-plugin
import { Load, useDefault } from '@dojo/core/load';

export function normalize(resourceId: string, resolver: (mid: string) => string): string {
    return resolver(mid).replace('app-name', 'appName');
}

export function load(mid: string, load: Load): Promise<any> {
    return load(mid).then(useDefault).then((value: string) => htmlToDom(string));
}
```

```
load('html-dom-plugin!path/to/html/string').then((htmlTree) => {
    // ...
});
```


docs/load.md
```diff
@@ -0,0 +1,108 @@
+# load
+
+## Module Exports
+
+### isPlugin - tests a value to determine whether is a plugin (an object with a `load` method).
```
dylans: whether it is a plugin? if not a sentence I wouldn't end with a period.

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+isPlugin([]); // false
+isPlugin([]); // false
+// false
+isPlugin({
+	observer() {}
+});
+
+```
+
+### load - dynamically loads a module or other resource.
```
dylans: Probably not end with a period.

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+});
+
+// Load a custom resource via a plugin
+load(require, 'plugin!./template.html').then(([ html ]: [ string ]) => {
+	// ...
+});
+
+```
+
+#### Using Load Plugins
```
dylans: While accurate, the grammar makes me think it should be Loader... so might adjust the wording or maybe cheat and say Load(er) ?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+
+The plugins that can be used with `load` loosely follow the [amdjs plugin API](https://github.com/amdjs/amdjs-api/blob/master/LoaderPlugins.md), with a few notable exceptions. Any plugin module MUST export a `load` method, and if the plugin has a default export, then the `load` method MUST exist on the default export. The plugin's `load` method will be passed a resource ID (the string that follows the `!` in the mid passed to `core/load`), as well as the `core/load` method itself. Unlike amdjs plugins, the plugin's `load` method is not passed a contextual `require` or a configuration object. Plugins can also expose an optional `normalize` method that is passed the resource ID and a resolver method (which will be either `require.toUrl`, `require.resolve`, or an identity function, depending on the environment). Note that if no `normalize` method is provided, then the provided resource ID will be resolved using `require.toUrl` or `require.resolve`, depending on the environment. Again, if the plugin module has a default export, the `normalize` method MUST exist on that object.
```
dylans: Might read better as an indented list of exceptions?

dylans: Should we just lowercase id since mid is lowercased also?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+```ts
+// Plugin that does not use the default export.
+import { Load } from 'src/load';
+
+export function normalize(resourceId: string, resolver: (id: string) => string): string {
+	return resolver(resourceId);
+}
+
+export function load(resourceId: string, load: Load) {
+	// This plugin does nothing more than load the resource ID with `core/load`.
```
dylans: id instead of ID ?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+// The same plugin, but using the default export
+import { Load } from 'src/load';
+
+const plugin = {
+	normalize(resourceId: string, resolver: (id: string) => string): string {
+		return resolver(resourceId);
+	},
+
+	load(resourceId: string, load: Load) {
+		// This plugin does nothing more than load the resource ID with `core/load`.
```
dylans: Maybe use resourceId for consistency?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+	}
+};
+export default plugin;
+
+```
+
+```ts
+import load from 'src/load';
+
+// 1. The module with the ID "plugin" is loaded.
```
dylans: I'd probably remove quotes around plugin, could use `plugin` perhaps?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+// 2. If "plugin" is not actually a plugin, then "plugin" itself is returned.
+// 3. If the plugin has a normalize method, then "some/resource/id" is passed to it,
+//    and the return value is used as the resource ID.
+// 4. The resource ID is passed to the plugin's `load` method.
```
dylans: resouceId ?

docs/load.md
```diff
@@ -0,0 +1,108 @@
...
+// 5. The loaded resource is used to resolve the load promise.
```
dylans: `load` instead of load?

src/load.ts
```diff
@@ -20,6 +20,38 @@ export interface Load {
 	(...moduleIds: string[]): Promise<any[]>;
 }
 
+export interface LoadPlugin<T> {
+	/**
+	 * An optional method that normmalizes a resource ID.
```
dylans: resourceId ?

src/load.ts
```diff
@@ -20,6 +20,38 @@ export interface Load {
...
+	 *
+	 * @param resourceId
+	 * The raw resource ID.
```
dylans: again, now sure if we should just call it resouceId here, or resource id

src/load.ts
```diff
@@ -20,6 +20,38 @@ export interface Load {
...
+	 *
+	 * @param resolver
+	 * A method that can resolve an ID to an absolute path. Depending on the environment, this will
```
dylans: `id` instead of ID?

src/load.ts
```diff
@@ -20,6 +20,38 @@ export interface Load {
...
+	 * usually be either `require.toUrl` or `require.resolve`.
+	 */
+	normalize?: (resourceId: string, resolver: (resourceId: string) => string) => string;
+
+	/**
+	 * A method that loads the specified resource.
+	 *
+	 * @param resourceId
+	 * The ID of the resource to load.
```
dylans: Same comment as earlier.

src/load.ts
```diff
@@ -38,34 +70,70 @@ export function useDefault(modules: any | any[]): any[] | any {
...
 				contextualRequire = require;
 			}
-			return new Promise(function (resolve) {
-				// TODO: Error path once https://github.com/dojo/loader/issues/14 is figured out
-				contextualRequire(moduleIds, function (...modules: any[]) {
-					resolve(modules);
+
+			return pluginLoad(moduleIds, load, (moduleIds: string[]) => {
+				return new Promise(function (resolve, reject) {
+					// TODO: Error path once https://github.com/dojo/loader/issues/14 is figured out
```
dylans: Given that this issue has been figured out, should something be changed here?

tests/unit/load.ts
```diff
@@ -45,6 +58,65 @@ const suite: any = {
 			assert.deepEqual(a, 'A');
 			assert.deepEqual(b, { 'default': 'B', three: 3, four: 4 });
 		});
+	},
+
+	'load plugin': {
+		'without a resource ID'(this: any) {
```
dylans: Same question about id vs ID vs resourceId in this section.

dylans (changes requested):
All minor stuff, looks like a really well done addition!

docs/load.md
mwistrand: Alternatively, it could be changed to just `Using Plugins` since it's already in the documentation for `load.ts`. Since `core/load` isn't really a loader, I tried to avoid using the word "loader" as much as possible.

src/load.ts
mwistrand: Probably `resourceId` for the variable name, and "resource id" for label.

docs/load.md
mwistrand: In my opinion, "resource id" works better since it's describing the value being passed in, as opposed to the variable.

src/load.ts
mwistrand: Yes, but doing so seems tangential to this particular PR. I'm more than happy to open another issue/PR specific to this, however.

dylans: Fair enough!

dylans (approved):

dylans (approved):
Code looks fine, only thought I had is that the error message for a failed test reads a lot like a successful test rather than a failure.

dylans (approved):

PR #276, opened 2017-01-23
Multimap implementation

### LINKED ISSUE #275: Provide a wrapper multi key map implementation
We have a use case for a Map that can consume an array as the key and internally maintain a map of maps for the array path.

Example usage:

```ts
map.set(['v', 'b', 'c', 'y'], 'b');
map.get(['v', 'b', 'c', 'y']); // returns `b`
```

agubler:
**Type:** feature

**Description:**

`MultiMap` implementation that supports an array of keys for a specific map value.

Resolves #275

### DIFF
--- a/src/MultiMap.ts
+++ b/src/MultiMap.ts
@@ -1,0 +1,156 @@
+import Symbol from '@dojo/shim/Symbol';
+import Map from '@dojo/shim/Map';
+import { ArrayLike } from '@dojo/shim/interfaces';
+import { forOf, Iterable, IterableIterator, ShimIterator } from '@dojo/shim/iterator';
+
+export default class MultiMap<T> {
+	private _map: Map<any, any>;
+	private _key: symbol;
+
+	constructor(iterable?: ArrayLike<[any[], any]> | Iterable<[any[], any]>) {
+		this._map = new Map<any, any>();
+		this._key = Symbol();
+		if (iterable) {
+			forOf(iterable, (value: [any[], any]) => {
+				this.set(value[0], value[1]);
+			});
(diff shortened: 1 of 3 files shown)


src/MultiMap.ts
```diff
@@ -0,0 +1,218 @@
...
+		});
+	}
+
+	/**
+	 * Deletes all keys and their associated values.
+	 */
+	clear(): void {
+		this._map.clear();
+	}
+}
```
rorticus, line 232: Since we're implementing other iterable thingies, do you think we should add a `Symbol.iterator` in here?

src/MultiMap.ts
```diff
@@ -0,0 +1,218 @@
...
+import Map from '@dojo/shim/Map';
+import { ArrayLike } from '@dojo/shim/interfaces';
+import { forOf, Iterable, IterableIterator, ShimIterator } from '@dojo/shim/iterator';
+
+/**
+ * A map implmentation that supports multiple keys for specific value.
+ *
+ * @param T Accepts the type of the value
+ */
+export default class MultiMap<T> {
```
rorticus: Out of curiosity, any reason we're not implementing or extending `Map` here?

rorticus (approved):

src/MultiMap.ts
agubler: A good question, I probably should `implement` rather than `extend` `Map` (and doing so means I need to add `size`)

src/MultiMap.ts
agubler: Yup no problemo

PR #282, opened 2017-02-03
Moving queue to dojo/shim and re-exporting it

rorticus:
Re-exporting queue from dojo/shim, as soon as https://github.com/dojo/shim/pull/72 lands. This also includes some small fixes to get core working with the latest in dojo/shim.

Note that the tests are going to fail until we get that dojo/shim PR merged and released.


package.json
```diff
@@ -22,7 +22,7 @@
   },
   "peerDependencies": {
     "@dojo/has": "2.0.0-alpha.7",
-    "@dojo/shim": "2.0.0-beta.8"
+    "@dojo/shim": "2.0.0-beta.9"
```
rorticus, line 25: ^ That version doesn't exist yet, but hopefully it will once we release shim 😃

dylans (approved):

agubler (approved):
Just re-exporting the ported code so 👍

PR #288, opened 2017-02-15
Adding default provider for node

rorticus:
Updating the request library so that it uses the node provider when running in node. This will mean we don't have to manually set the provider every time we want to use the request library.

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -43,7 +43,7 @@ Object.defineProperty(request, 'setDefaultProvider', {
 
 providerRegistry.setDefaultProvider(xhr);
 
-if (!has('host-node')) {
+if (has('host-node')) {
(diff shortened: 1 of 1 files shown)


src/request.ts
```diff
@@ -42,6 +43,11 @@ Object.defineProperty(request, 'setDefaultProvider', {
 
 providerRegistry.setDefaultProvider(xhr);
 
+if (!has('host-browser')) {
```
dylans: While this is probably fine, I'd prefer an explicit feature test for Node if possible. I know that we currently only focus on browsers or Node, but a lot can change in the future.

rorticus: Changed to check for `host-node`

matt-gadd (approved):

PR #292, opened 2017-02-23
deep assign/mixin through nested object keys

### LINKED ISSUE #290: Confusion about lang.deepMixin and other deep... (urgent)
Preclaimer: What I need is the old `deepCopy` which is currently

> "in an odd location"

in dojo1, see  http://dojo-toolkit.33424.n3.nabble.com/deep-lang-mixin-td4007082.html 
but where is it in dojo2 ???

`lang/deepMixin` and `lang/deepAssign` behave totally different from `deepCopy` 
(same behaviour like plain /mixin for me), 
check Dylans example https://jsfiddle.net/dylan/oja45dv8/ 

I would expect lang/deepMixin to return the same than the old deepCopy or did I overlook sth. ?

agubler:
**Type:** bug

**Description:**

Addresses an issue with deep assigning and deep mixing objects where nested objects where not merged with target but overrode the target.

Resolves #290

### DIFF
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -48,6 +48,7 @@ function _mixin<T extends {}, U extends {}>(kwArgs: MixinArgs<T, U>): T&U {
 	const inherited = kwArgs.inherited;
 	const target: any = kwArgs.target;
 	const copied = kwArgs.copied || [];
+	const copiedClone = [ ...copied ];
 
 	for (let source of kwArgs.sources) {
 		if (source === null || source === undefined) {
@@ -57,7 +58,7 @@ function _mixin<T extends {}, U extends {}>(kwArgs: MixinArgs<T, U>): T&U {
 			if (inherited || hasOwnProperty.call(source, key)) {
 				let value: any = source[key];
 
-				if (copied.indexOf(value) !== -1) {
+				if (copiedClone.indexOf(value) !== -1) {
 					continue;
 				}
 
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -40,12 +40,14 @@ interface MixinArgs<T extends {}, U extends {}> {
 	inherited: boolean;
 	sources: (U | null | undefined)[];
 	target: T;
+	copied?: any[];
 }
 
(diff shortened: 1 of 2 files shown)


src/lang.ts
```diff
@@ -40,36 +40,44 @@ interface MixinArgs<T extends {}, U extends {}> {
...
 	for (let source of kwArgs.sources) {
 		if (source === null || source === undefined) {
 			continue;
 		}
 		for (let key in source) {
 			if (inherited || hasOwnProperty.call(source, key)) {
-				let value: any = (<any> source)[key];
+				let value: any = source[key];
+
+				if (copied.indexOf(value) !== -1) {
```
bryanforbes: Won't this skip assignment if the same object is referenced by two properties of the source object?

```ts
const a = { foo: 1 };
const b = { one: a, two: a };
const c = deepAssign({}, b);
```

agubler: @bryanforbes you are correct, I think I have now addressed this.

agubler:
@pottedmeat I believe this addresses the concerns we had and hopefully still guards against circular references. Please let me know if you can think of any more scenarios that are not covered by the unit tests.

agubler:
@kitsonk do you think you could give this a review? I believe it covers all the scenarios whilst also guarding against cyclical references.

src/lang.ts
```diff
@@ -40,36 +40,45 @@ interface MixinArgs<T extends {}, U extends {}> {
...
+				if (copiedClone.indexOf(value) !== -1) {
+					continue;
+				}
 
 				if (deep) {
 					if (Array.isArray(value)) {
 						value = copyArray(value, inherited);
 					}
 					else if (shouldDeepCopyObject(value)) {
+						const targetValue: any = target ? target[key] || {} : {};
```
kitsonk: appears there is only partial coverage on this line

kitsonk (commented):
Just one note.

kitsonk (approved):
👍

PR #305, opened 2017-03-08
Hotfix: move common load methods into load/util.

### LINKED ISSUE #304: `load/webpack` should not import from `load`.
**Bug**

Previously when building Dojo 2 applications, calls to `load` would be replaced by webpack to calls to `load/webpack`, with methods from `load` being re-exported as needed. This no longer works, so any methods common to both `load` and `load/webpack` need to be separated into a dedicated module.

mwistrand:
**Type:** bug

**Description:**

`src/load/webpack` is no longer including the values imported from `src/load` in webpack builds. Separating the common methods between the two into a shared module circumvents this issue.

Resolves #304

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -1,6 +1,6 @@
-import { forOf, isIterable, isArrayLike } from '@dojo/shim/iterator';
 import Promise from '@dojo/shim/Promise';
 import { Require } from '@dojo/interfaces/loader';
+import { isPlugin } from './load/util';
 
 declare const require: Require;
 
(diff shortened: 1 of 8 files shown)


mwistrand:
One question: should `load` re-export `isPlugin` and `useDefault` since they were originally exports of `load`, or is it fine to require others to import them from `load/util` if needed?

agubler:
@mwistrand we should re-export those functions from `loads.

matt-gadd (approved):

PR #315, opened 2017-04-07
Fix for binary data getting corrupted in node request provider

rorticus:
**Type:** bug / feature

**Description:**

Binary data (a zip file in this case) was being corrupted in the node request provider because it would convert everything to a string and that was causing some issues.

Also explicitly ignoring the `None` content encoding.

### DIFF
--- a/src/request/providers/node.ts
+++ b/src/request/providers/node.ts
@@ -468,7 +468,7 @@ export default function node(url: string, options: NodeRequestOptions = {}): Tas
 							if (contentEncodings.length) {
 								const encoding = contentEncodings.pop()!.trim();
 
-								if (encoding === '' || encoding === 'identity') {
+								if (encoding === '' || encoding.toLowerCase() === 'none' || encoding === 'identity') {
 									// do nothing, response stream is as-is
 									handleEncoding();
 								}
--- a/src/request/providers/node.ts
+++ b/src/request/providers/node.ts
@@ -88,7 +88,7 @@ interface HttpsOptions extends Options {
 interface RequestData {
 	task: Task<http.IncomingMessage>;
 	buffer: any[];
-	data: string;
+	data: Buffer | string;
 	size: number;
 	used: boolean;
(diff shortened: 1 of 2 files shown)


src/request/providers/node.ts
```diff
@@ -473,7 +468,7 @@ export default function node(url: string, options: NodeRequestOptions = {}): Tas
 							if (contentEncodings.length) {
 								const encoding = contentEncodings.pop()!.trim();
 
-								if (encoding === '' || encoding === 'identity') {
+								if (encoding === '' || encoding.toLowerCase() === 'none' || encoding === 'identity') {
```
jason0x43: Instead of lower-casing one specific case here, maybe we should lowercase the encoding as it's popped off `contentEncodings`.

src/request/providers/node.ts
```diff
@@ -88,7 +88,7 @@ interface HttpsOptions extends Options {
 interface RequestData {
 	task: Task<http.IncomingMessage>;
 	buffer: any[];
-	data: string;
+	data: Buffer | string;
```
jason0x43: Since data should only ever be a buffer (the only thing that's assigned to it is `dataAsBuffer`), should its type here be Buffer? The initializer would then need to change to use something like `Buffer.alloc(0)`.

jason0x43 (commented):
Overall this looks fine. I had a couple of suggestions, but they're internal (they won't affect the usage of the module).

src/request/providers/node.ts
rorticus: I guess that would be alright wouldn't it. I was afraid of lowercasing them all and changing the current behavior but since everything being checked for now is lowercased anyway, I think its ok.

src/request/providers/node.ts
rorticus: Good point. This has been updated!

rorticus:
CI build seems to have failed due to unrelated docs issue.

jason0x43 (approved):

PR #317, opened 2017-04-09
Enhance `assign()` typings to allow different types of sources.

### LINKED ISSUE #291: Use Partial for the assign/mixin typings
The sources for assign/mixin functions should be typed so that subsequent sources can contain attributes that are not included in the first `source`. At the moment compile errors occurs if the second source expands the keys compared to the initial `source`.

Code:
```ts
const baseObject = {
	foo: 'bar'
};

lang.assign({}, baseObject, { baz: 'foo' });
```

Expected:

Compile successfully

Actual:

`error TS2453: The type argument for type parameter 'U' cannot be inferred from the usage. Consider specifying the type arguments explicitly.
  Type argument candidate '{ foo: string; }' is not a valid type argument because it is not a supertype of candidate '{ baz: string; }'.
    Object literal may only specify known properties, and 'baz' does not exist in type '{ foo: string; }'.`

lzhoucs:
**Type:** bug

**Description:**

The updated typings for `assign()` is the same as in [Typescript](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es6.d.ts#L4461-L4480).

I looked at `Partial` but so far I haven't found a way to solve the problem with it. 

Although this solution should resolve #291 , it is still based on the same intersection types we've been using, so due to the limitation of intersection types, there's still other typings issues. For example:
```ts
const result  = lang.assign({foo: 1}, {foo: 'foo'}); // result.foo becomes `string & number`
result.foo = 'bar'; // won't compile
```
And a workaround:
```ts
// workaround
type Foo = { foo: string };
const result: Foo  = lang.assign({foo: 1}, {foo: 'foo'}); // result.foo becomes `string`
result.foo = 'bar'; // works
```

Resolves #291

### DIFF
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -87,7 +87,10 @@ function _mixin<T extends {}, U extends {}>(kwArgs: MixinArgs<T, U>): T&U {
 }
 
 interface ObjectAssignConstructor extends ObjectConstructor {
-	assign<T extends {}, U extends {}>(target: T, ...sources: (U | null | undefined)[]): T&U;
+	assign<T, U>(target: T, source: U): T & U;
+	assign<T, U1, U2>(target: T, source1: U1, source2: U2): T & U1 & U2;
+	assign<T, U1, U2, U3>(target: T, source1: U1, source2: U2, source3: U3): T & U1 & U2 & U3;
+	assign(target: any, ...sources: any[]): any;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


agubler (approved):
Seems like a reasonable approach to at least mirror the TS typings for `Object.assign`, the CI failure is looks typedoc related.

kitsonk:
Merged this, because the CI is passing except for the api document generating trying to be generated on non-master branches... 😢

PR #319, opened 2017-04-11
Has test for `blob` should use a valid URI

### LINKED ISSUE #318: Invalid URL for blob has sniff
**Bug** 

The URI of `/` is not valid for the blob sniff in `has.ts`, which causes a failure running tests for projects built with webpack.

This should be updated to `/foo` (or another valid URI)

agubler:
**Type:** bug

**Description:**

Update `has` check for `blob` to a valid URI.

Resolves #318

### DIFF
--- a/src/has.ts
+++ b/src/has.ts
@@ -17,7 +17,7 @@ add('blob', function () {
 	}
 
 	const request = new XMLHttpRequest();
-	request.open('GET', '/', true);
+	request.open('GET', '/foo', true);
 	request.responseType = 'blob';
 	request.abort();
 	return request.responseType === 'blob';


matt-gadd (approved):

agubler:
This is failing for typedoc, all other CI checks passed. Merging.

msssk:
This should use an absolute URI, see https://github.com/dojo/dojo/pull/260.

It should also include a unit test like the one the PR linked above.

agubler:
@msssk You're correct, it still doesn't work with `/foo` in a jsdom environment (i do not understand why it was working locally for me yesterday)

PR #323, opened 2017-04-20
fixed usage of node auth with special characters

devpaul:
**Type:** bug

request/provider/node was url encoding user names and passwords for basic authentication. Node's http/s `request()` handles this already by doing a base64 encoding. This is breaking usernames and passwords for Basic auth with special characters due to the extra url encoding.

### DIFF
--- a/src/request/providers/node.ts
+++ b/src/request/providers/node.ts
@@ -242,7 +242,7 @@ export function getAuth(proxyAuth: string | undefined, options: NodeRequestOptio
 	}
 
 	if (options.user || options.password) {
-		return encodeURIComponent(options.user || '') + ':' + encodeURIComponent(options.password || '');
+		return `${ options.user || '' }:${ options.password || '' }`;
 	}
 
 	return undefined;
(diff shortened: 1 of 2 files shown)


devpaul:
In the meantime, if anyone needs a workaround for this:

```typescript
import { RequestOptions, Response } from '@dojo/core/request/interfaces';
import request, { Provider } from '@dojo/core/request';
import nodeRequest from '@dojo/core/request/providers/node';
import Task from '@dojo/core/async/Task';

export function fixAuth(options: RequestOptions) {
	if (options.password || options.user) {
		const credentials = new Buffer(`${ options.user || '' }:${ options.password || ''}`).toString('base64');
		const headers = options.headers = <{ [key: string]: string }> options.headers || {};
		headers['Authorization'] = `Basic ${ credentials }`;
		delete options.password;
		delete options.user;
	}
	return options;
}

const provider: Provider = function (url: string, options: RequestOptions): Task<Response> {
	return nodeRequest(url, fixAuth(options));
};

request.setDefaultProvider(provider);

export default request;
```

dylans (approved):

kitsonk (commented):
Was this tied to any particular version of Node.js.  It seems odd we would have been doing something (though untested) only to have it break later on.  If there was a change in Node.js, we should identify it so we make sure we have the right compatibility expectations.

devpaul:
Nothing came up in a search of issues from nodejs/node that looked like it would apply.

If authentication was done in the URL rather than as an auth header then it's possible that you'd want to url encode the values, but then you wouldn't base64 encode it.

kitsonk (approved):

PR #326, opened 2017-04-28
Adding upload progress reporting, issue #285

### LINKED ISSUE #285: Add upload progress reporting
Once  https://github.com/dojo/core/pull/261 lands, we should add upload progress reporting to the xhr and node providers. Currently, they only provide reporting on download progress.

rorticus:
**Type:** feature

**Description:**

Adds upload progress monitoring to the node & XHR request providers. It is implemented as a request option because by the time the returned promise is resolved, the uploads have already finished. Basic usage is,

``` typescript
const uploadObserver = new UploadObserver();
uploadObserver.on('upload', event => {
    console.log(`Total bytes uploaded: ${event.totalBytesUploaded}`);
});
request.get('http://www.example.com', {
    method: 'POST',
    body: aLotOfData,
    uploadObserver
});
```

`XHR` was easy as it emits upload events already. The node uploader was a bit more tricky, as there doesn't seem to be a way to find out how much data node has uploaded when using `request.write`. As such, I've added support for passing a readable stream to the node provider. As data is read from the stream, it is written to the request, and upload progress can be reported.

Resolves #285


rorticus:
CI failures seem to be an unrelated issue 😄

docs/request.md
```diff
@@ -9,3 +9,39 @@ response.
...
+uploader.on('upload', uploadEvent => {
+	console.log(`Total uploaded: ${uploadEvent.totalBytesUploaded}`);
+});
+request.post('http://www.example.com/', {
+	body: someLargeString,
+	uploadObserver: uploader
+});
+```
+
+Note that while the node provider will emit a single `upload` event when it is done uploading, it cannot emit more granular upload events with `string` or `Buffer` body types. To receive more frequent upload events, you can use the `bodyStream` option to provide a `Readable` with the body content. Upload events will be emitted as the data is read from the stream.
```
dylans: node -> Node.js

dylans (changes requested):
one minor nit from me, this looks reasonable though I'm sure someone else should also review this.

src/request.ts
```diff
@@ -35,6 +35,16 @@ const request: {
 	});
 });
 
+[ 'POST', 'PUT' ].forEach(method => {
+	Object.defineProperty(request, method.toLowerCase(), {
+		value(url: string, options: RequestOptions = {}): UploadObservableTask<Response> {
```
kitsonk, line 40: We don't have a test for request options default to nothing, therefore we are missing a branch in our testing.

src/request/SubscriptionPool.ts
```diff
@@ -0,0 +1,25 @@
+import { SubscriptionObserver } from '@dojo/shim/Observable';
+
+export default class SubscriptionPool<T> {
+	private _observers: SubscriptionObserver<T>[] = [];
+
+	add(subscription: SubscriptionObserver<T>) {
+		this._observers.push(subscription);
+
+		return () => {
+			this._observers.splice(this._observers.indexOf(subscription), 1);
```
kitsonk: Not being covered in the tests.

src/request/SubscriptionPool.ts
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	next(value: T) {
+		this._observers.forEach(observer => {
+			observer.next(value);
+		});
+	}
+
+	complete() {
+		this._observers.forEach(observer => {
```
kitsonk, line 40: Not covered in testing.

src/request/providers/node.ts
```diff
@@ -10,10 +10,13 @@ import { deepAssign } from '../../lang';
 import { queueTask } from '../../queue';
 import { createTimer } from '../../util';
 import Headers from '../Headers';
-import { RequestOptions } from '../interfaces';
+import { RequestOptions, UploadObservableTask } from '../interfaces';
 import Response from '../Response';
 import TimeoutError from '../TimeoutError';
 import { Readable } from 'stream';
+import { UploadObserverEvent } from '../interfaces';
+import Observable from '../../Observable';
```
kitsonk: Codecov says this file isn't covered at all.

src/request.ts
```diff
@@ -35,6 +35,16 @@ const request: {
...
+			options = Object.create(options);
+			options.method = method;
+			return <UploadObservableTask<Response>> request(url, options);
```
kitsonk, line 43: While it isn't a big deal, we are trying to move to `return request(url, options) as UploadObservableTask<Response>;`  The TypeScript team is recommending it to avoid issues with JSX.  Though we don't plan to use JSX, it still feels cleaner/more readable.

kitsonk (changes requested):
A few comments.

rorticus:
@dylans Updated to use `Node.js` 👍

rorticus:
@kitsonk OK, I've refactored to use Observables for everything. See the README for details, but basically `Request` exposes `upload` and `Response` exposes `download` and `data`.

src/request/SubscriptionPool.ts
kitsonk: Fixed now

src/request/SubscriptionPool.ts
```diff
@@ -0,0 +1,40 @@
...
+		};
+	}
+
+	next(value: T) {
+		if (this._observers.length === 0) {
+			this._queue.push(value);
+
+			// when the queue is full, get rid of the first ones
+			while (this._queue.length > this._queueMaxLength) {
+				this._queue.shift();
```
kitsonk, line 30: Not covered in testing.

kitsonk (approved):

kitsonk (changes requested):
We should cover the missing line and branches.

rorticus:
@kitsonk Added a few more tests for better coverage

dylans (approved):
My feedback has been addressed... obviously Kit needs to verify that his feedback has as well before this is landed.

rorticus:
We might be able to get rid of `QueuingEvented`, but I'm not sure if anyone is using it... It's no longer being used in `core` but it looks like it's been modified since creation, so I just left it alone.

kitsonk:
The only thing I am missing is the full coverage on [this line](https://github.com/dojo/core/pull/326/files#diff-31b74031f3c83e9e883fd930aaf87f54R40).  If it isn't testable, then we don't need the default argument.

rorticus:
@kitsonk That line should be covered now!

kitsonk (approved):
Happy camper now! 😂 ⛺️

PR #327, opened 2017-04-28
Only do "grunt doc" on master

kitsonk:
**Type:** bug

**Description:**

This skips the `grunt doc` when not building on master, so our PRs can build without issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ script:
 - grunt intern:saucelabs --combined
 - grunt remapIstanbul:ci
 - grunt uploadCoverage
-- grunt doc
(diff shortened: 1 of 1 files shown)


devpaul:
I removed the DEPLOY_DOCS environment variable, which is what triggers the publish. I'm OK w/ this change if you feel it's necessary to disable building API docs because typedoc is problematic right now.

devpaul (approved):
I removed the DEPLOY_DOCS environment variable, which is what triggers the publish. I'm OK w/ this change if you feel it's necessary to disable building API docs because typedoc is problematic right now.

agubler (approved):

PR #330, opened 2017-05-06
Support symbols for event type

### LINKED ISSUE #329: Support `Evented` with symbol types
**Enhancement** 

Add support to use a `Symbol` as an event type.

Package Version: `beta1`

**Code**

```ts
const myEventType = Symbol();
const evented = new Evented({});

evented.on(myEventType, () => {});
```

```ts
const myEventType = Symbol();
const evented = new Evented({});

evented.emit({ type: myEventType });
```

**Expected behavior:**

No errors

**Actual behavior:**

Errors as Symbols are not supported.

agubler:
**Type:** feature

**Description:**

Adds support to use symbols as an event type.

As part of this PR I have moved interfaces from `@dojo/interfaces` that have been affected into the `Evented` module; It feels cumbersome to need to make interface changes in `@dojo/interfaces` (that could be breaking, which is pretty much unmanageable) and then release `@dojo/interfaces` in order to get make fixes/enhancements downstream.

Resolves #329

### DIFF
--- a/src/Evented.ts
+++ b/src/Evented.ts
@@ -1,6 +1,6 @@
 import { Actionable } from '@dojo/interfaces/abilities';
-import { EventedListener, EventedListenerOrArray, EventedListenersMap, EventedCallback } from '@dojo/interfaces/bases';
-import { EventObject, EventTargettedObject, EventErrorObject, Handle } from '@dojo/interfaces/core';
+import { EventedListener, EventedListenerOrArray, EventedListenersMap } from '@dojo/interfaces/bases';
+import { EventTargettedObject, EventErrorObject, Handle } from '@dojo/interfaces/core';
 import Map from '@dojo/shim/Map';
 import { on as aspectOn } from './aspect';
 import { Destroyable } from './Destroyable';
@@ -36,6 +36,22 @@ function handlesArraytoHandle(handles: Handle[]): Handle {
 	};
 }
 
+export interface EventObject {
+	/**
+	 * The type of the event
+	 */
+	readonly type: string | symbol;
+}
+
+export interface EventedCallback<E extends EventObject> {
+	/**
+	 * A callback that takes an `event` argument
+	 *
+	 * @param event The event object
+	 */
+	(event: E): boolean | void;
+}
+
 /**
  * Interface for Evented constructor options
  */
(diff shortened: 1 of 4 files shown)


agubler:
Needs some js documentation.

src/Evented.ts
```diff
@@ -60,6 +76,8 @@ export interface BaseEventedEvents {
 	 */
 	(type: string, listener: EventedListenerOrArray<Evented, EventTargettedObject<Evented>>): Handle;
 
+	(type: symbol, listener: EventedListenerOrArray<Evented, EventTargettedObject<Evented>>): Handle;
```
agubler: can probably go in the signature above as `string | symbol`.

src/Evented.ts
```diff
@@ -36,6 +36,22 @@ function handlesArraytoHandle(handles: Handle[]): Handle {
...
 }
 
+export interface EventObject {
+	/**
+	 * The type of the event
+	 */
+	readonly type: string | symbol;
+}
+
+export interface EventedCallback<E extends EventObject> {
```
agubler, line 49: At least take the documentation from interfaces for both the interfaces.

src/Evented.ts
```diff
@@ -36,6 +36,22 @@ function handlesArraytoHandle(handles: Handle[]): Handle {
 	};
 }
 
+export interface EventObject {
+	/**
+	 * The type of the event
+	 */
+	readonly type: string | symbol;
```
kitsonk, line 46: Are we going to get into assignability issues with some code taking `EventObject` from `@dojo/interfaces/core` and some coming from this with a `symbol` in there?

agubler: A good question, I would have tested it but I cannot find any usages of either `EventObject` or `EventedCallback` outside of `Evented` (which suggests that both these interfaces should actually be in the `Evented` package).

agubler: However there are other packages in `core` that use the interfaces `EventObject` (which I haven't updated) and does not cause any issues.

kitsonk: `@dojo/core/on` uses it, but it seems like everyone else doesn't specifically refer to that interface directly when writing overloads for other `.on()` methods for custom events.  Historically, it had cross cutting concerns when `on()` and `Evented` were in separate packages.

agubler: I've updated the internal references to `EventObject` to use the local interface.

kitsonk (approved):

PR #339, opened 2017-05-26
Adding base64 encoding from web-editor and removed old encoding

### LINKED ISSUE #299: dojo/core encoding.ts - base64 does not work anymore
I am not sure about base64 but it returns different results compared to my own base64 functions ...
However even directly decoding an encoded string returns a wrong result :
A very simple test was
console.log( base64.decode(base64.encode('Lorem')) );

Should result: 'Lorem'
This results in 'Lore' ...

Apart from that if I've got characters like `+` it was a completely different result.

What I am basically looking for are the functions to encode/decode any string to base64 and to base64Url. This is what I am currently using:
```Typescript
export function escape(str: string) {
  return str.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '');
}
export function unescape(str: string) {
  str = str.toString();
  var mod = str.length % 4;
  if (mod !== 0) { str += _repeat('=', 4 - mod); }
  return str.replace(/\-/g, '+').replace(/_/g, '/');
}

function _repeat(str: string, num: number) { return new Array(num + 1).join(str); };
/* base64 --> */
export function base64Encode(str: string) {
  if (has('host-node')) {
    return new Buffer(str).toString('base64');
  } else {
    return btoa(encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, function(m, p1) {
      return String.fromCharCode(parseInt(('0x'+p1), 16));
    }));
  }
}
export function base64Decode(b64String: string | Buffer) {
  b64String = b64String.toString();
  if (has('host-node')) {
    return (new Buffer(b64String, 'base64')).toString('utf8');
  } else {
    return decodeURIComponent(b64String.split('').map(function(c) {
        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2);
    }).join(''));
  }
}
export function base64UrlEncode(str: string): string {
  return escape(base64Encode(str));
}
export function base64UrlDecode(b64String: string | Buffer): string {
  b64String = b64String.toString();
  return base64Decode(unescape(b64String));
}
/* <-- base64 */
```

rorticus:
**Type:** bug / feature

**Description:**

Removing old `encoding.ts` and replaced it with the base64 methods from `web-editor`. It doesn't look like the code from `encoding.ts` was being used, so I think we can just ditch it.

Resolves #299


kitsonk (approved):
The only thing is I would want to get someone else's opinion on naming of the module...  I don't trust my own judgement.  (Someone else should review this too since it is porting of code I wrote)

src/base64.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Take a string encoded in base64 and decode it
+ * @param encodedString The base64 encoded string
+ */
+export const decode: (encodedString: string) => string = has('atob') ? function (encodedString: string) {
+	/* this allows for utf8 characters to be decoded properly */
+	return decodeURIComponent(Array.prototype.map.call(atob(encodedString), (char: string) => '%' + ('00' + char.charCodeAt(0).toString(16)).slice(-2)).join(''));
+} : function (encodedString: string): string {
+	return new Buffer(encodedString.toString(), 'base64').toString('utf8');
+};
```
mwistrand, line 16: Essentially we are assuming that all supported browsers support `btob`/`atob`, since the alternative is the Node-specific `Buffer`?

rorticus: @mwistrand yeah, and I think we are good on that account, https://caniuse.com/#search=atob . Supported back to IE 10.

mwistrand (approved):

PR #346, opened 2017-07-07
Changes to align with shim refactors

kitsonk:
**Type:** feature

**Description:**

These PRs are updates to keep in sync with changes in PR dojo/shim#101


src/aspect.ts
```diff
@@ -281,7 +281,7 @@ function getDispatcherObject(target: Targetable, methodName: string): Dispatcher
 function getJoinPointDispatcher<F extends GenericFunction<T>, T>(joinPoint: F): F {
 
 	function dispatcher(this: Function, ...args: any[]): T {
-		const { before, after, joinPoint } = dispatchAdviceMap.get(dispatcher);
+		const { before, after, joinPoint } = dispatchAdviceMap.get(dispatcher)!;
```
agubler: I don't think we should be ignoring the strict null checks, I believe you mentioned that it can now correctly return `undefined` so we should probably guard against that appropriately.

kitsonk: I was debating this, since from the code below, you could never get a situation where the map for the `dispatcher` is `undefined`, because the only way the way the function could be called is if its map had been set.

rorticus (approved):

src/aspect.ts
kitsonk: I have added comments to document the use of not null assertion is appropriate because we can't logically have `undefined` values from the maps.

agubler (approved):

PR #350, opened 2017-07-31
Adding a flag to disable automatic inclusion of x-requested-with

### LINKED ISSUE #328: request provider always sets X-Requested-With
**Bug**

While it is helpful to have `xhr` set the header `X-Requested-With` as it is a common case for CORS that fails because the end user doesn't know it is often required, if the server you are facing off doesn't allow the header to be specified in the `OPTIONS` preflight in that is performed in Chromium based browsers, the request fails with the message `Request header field X-Requested-With is not allowed by Access-Control-Allow-Headers in preflight response.` and there is no way to turn the header off.

Package Version: beta1

rorticus:
**Type:** feature

**Description:**

Adding an XHR request option, `includeRequestedWith` that defaults to `true`, that can be used to disable the automatic inclusion of the `X-Requested-With`.

Example,

```typescript
xhr('some-url', { includeRequestedWith: false }).then(...)
```

Resolves #328

### DIFF
--- a/src/request/providers/xhr.ts
+++ b/src/request/providers/xhr.ts
@@ -18,6 +18,7 @@ import SubscriptionPool from '../SubscriptionPool';
  */
 export interface XhrRequestOptions extends RequestOptions {
 	blockMainThread?: boolean;
+	includeRequestedWithHeader?: boolean;
 }
 
 interface RequestData {
@@ -270,6 +271,7 @@ export default function xhr(url: string, options: XhrRequestOptions = {}): Uploa
 
 	let hasContentTypeHeader = false;
 	let hasRequestedWithHeader = false;
+	const { includeRequestedWithHeader = true } = options;
 
 	if (options.headers) {
 		const requestHeaders = new Headers(options.headers);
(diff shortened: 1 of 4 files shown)


agubler (approved):

src/request/providers/xhr.ts
```diff
@@ -18,6 +18,7 @@ import SubscriptionPool from '../SubscriptionPool';
  */
 export interface XhrRequestOptions extends RequestOptions {
 	blockMainThread?: boolean;
+	includeRequestedWithHeader?: boolean;
```
kitsonk: My only concern with this is that in general we don't have good documentation for any of this, and now we are adding to it.  Without any JSDoc, the API docs won't contain anything as well.  At the very least, we should add JSDoc for options we add, we already have a cleanup task to go back over and review stuff, so lets just not add to the pile of debt.

agubler: That's a good point Kit, we should definitely be looking to at least have JSDoc (and even better README doc too)

kitsonk: Yeah, I was going to say README, but realise for core that is a ginormous task and actually for lots of it, just sane API docs is likely to be better, maybe...  Adding this particular flag to the README would be a bit crazy as it stands at the moment, since this is our `request` documentation at the moment:

> ### HTTP requests
>
> The `@dojo/core/request` module contains methods to simplify making http requests. It can handle making requests in both node and the browser through the same methods.

rorticus: Added some jsdoc for request options 😄

kitsonk (approved):

PR #356, opened 2017-10-11
Update tests to Intern 4

### LINKED ISSUE #347: Update tests to Intern 4

bryanforbes:
**Type:** enhancement

**Description:**

This PR updates the unit and functional tests from Intern 3 to Intern 4

Resolves #347


intern.json
```diff
@@ -0,0 +1,117 @@
...
+			"capabilities+": {
+				"browserstack.debug": false
+			},
+
+			"environments+": [
+				{ "browserName": "internet explorer", "version": "11" },
+				{ "browserName": "edge" },
+				{ "browserName": "firefox", "platform": "WINDOWS" },
+				{ "browserName": "chrome", "platform": "WINDOWS" }
+				// Issue with iphone 9.1 and Safari compatability, commented out for move to
```
agubler: comments in a `json` file?

bryanforbes: Isn't it wonderful?

tests/functional/text/textPlugin.ts
```diff
@@ -1,49 +1,42 @@
...
-		.then(pollUntil<any>(function() {
-			return (<any> window).loaderTestResults;
-		}, undefined, timeout), undefined)
-		.then(testFn, function() {
-			throw new Error('loaderTestResult was not set.');
-		});
+async function executeTest(suite: Suite, htmlTestPath: string, timeout = 10000) {
+	try {
+		return await suite.remote.get(htmlTestPath).then(pollUntil(function () {
+			return (<any> window).loaderTestResults || null;
```
agubler, line 12: would it be better to use `dojo/shim/global`?

tests/unit/DateObject.ts
```diff
@@ -51,169 +49,171 @@ registerSuite({
...
-			object = new DateObject({ year: 1995, month: 12, dayOfMonth: 31 });
-			object.year += 35;
-			assert.strictEqual(+object, +new Date(2030, 11, 31));
-		},
-
-		'month': function () {
-			assert.strictEqual(object.month, date.getMonth() + 1);
-			date.setMonth((object.month = 1) - 1);
-			assert.strictEqual(object.month, date.getMonth() + 1);
+		tests: {
```
agubler, line 52: what's the purpose of the `test` wrapping object?

tests/unit/Destroyable.ts
```diff
@@ -1,9 +1,8 @@
-import * as registerSuite from 'intern!object';
-import * as assert from 'intern/chai!assert';
+const { registerSuite } = intern.getInterface('object');
+const { assert } = intern.getPlugin('chai');
 import { Destroyable } from '../../src/Destroyable';
 
-registerSuite({
-	name: 'bases/Destroyable',
+registerSuite('bases/Destroyable', {
```
agubler: do we need `bases` in the name now? Think this is a hangover from where they were moved from (compose)

tests/unit/MultiMap.ts
```diff
@@ -75,16 +73,18 @@ registerSuite({
...
 		},
 
-		'key found'() {
-			assert.strictEqual(map.get([ 1, 2 ]), 'bar');
-			assert.isTrue(map.delete([ 1, 2 ]));
-			assert.isUndefined(map.get([ 1, 2 ]));
-			assert.strictEqual(map.get([ 1 ] ), 'foo');
-			assert.strictEqual(map.get([ 2 ] ), 'baz');
-		},
+		tests: {
```
agubler, line 76: tests wrapping object again?

tests/unit/QueuingEvented.ts
```diff
@@ -8,9 +8,7 @@ interface CustomEvent {
 	value: number;
 }
 
-registerSuite({
-		name: 'Evented',
-
+registerSuite('Evented', {
```
agubler: QueingEvented

typings.json
```diff
@@ -1,6 +1,3 @@
 {
```
agubler, line 3: can remove the file if there are no more typings required.

tsconfig.json
```diff
@@ -14,6 +14,12 @@
 		"moduleResolution": "node",
 		"noImplicitAny": true,
 		"noImplicitThis": true,
+		"plugins": [
```
agubler: what's this?

agubler (commented):
some newbie questions a couple of which have already been answered :)

tests/unit/DateObject.ts
agubler: I've found out, that if there are lifecycle functions such as `before` then all tests need to be wrapped in a `tests` object.

tests/unit/MultiMap.ts
agubler: ignore, answered [here](https://github.com/dojo/core/pull/356#discussion_r144129424)

intern.json
agubler: Erm not especially, what is intern using to allow comments? Perhaps we should remove them.

tests/functional/text/textPlugin.ts
bryanforbes: This function is always run on the client and it would be a big overhead to `require()` the shim for this simple polling function.

tests/unit/Destroyable.ts
bryanforbes: If you'd like me to update this as part of the PR, just give me the word

tests/functional/text/textPlugin.ts
agubler: is it a big overhead to require a single module from shim?

tests/unit/Destroyable.ts
agubler: as you're moving the name might be nice 👍

tsconfig.json
bryanforbes: It loads a tslint plugin so linting errors come through in the language services. We're using it for Intern 4 and makes the development experience quite nice.

agubler: Okay. We should probably put an issue in meta if we want to add that to the Dojo 2 packages.

agubler: rather than include it in core as part of the intern 4 conversion.

tests/functional/text/textPlugin.ts
bryanforbes: Since this is a functional test, the text of this function is sent to the client and run synchronously. There's no way to do a `require(['@dojo/shim/global'], function () { ... })` in the execution on the client for a `pollUntil`.

Gruntfile.js
```diff
@@ -9,6 +9,9 @@ module.exports = function (grunt) {
 			options: {
 				ignoreCompilerErrors: true, // Remove this once compile errors are resolved
 			}
+		},
+		intern: {
+			version: 4
```
rorticus, line 14: We don't need this anymore with your new grunt-dojo2 changes, right?

rorticus (approved):

Gruntfile.js
bryanforbes: `grunt-dojo2` still defaults to Intern 3 for backwards compatibility, so we need this

kitsonk (approved):
No comments that were already raised.

Issue #357, opened 2017-10-16
`utils#throttle()` test fails in Edge


bryanforbes:
**Bug**

When running the unit tests on Edge, the following error is raised:

```
× MicrosoftEdge on windows 10 - utility functions - throttle - throttles callback (0.011s)
    AssertionError: Function should not be called until throttle delay has elapsed: expected 7 to be above 23
      at r  <node_modules/intern/browser/remote.js:69:491>
      at n.prototype.assert  <node_modules/intern/browser/remote.js:445:1237>
      at c  <node_modules/intern/browser/remote.js:458:2068>
      at s  <node_modules/intern/browser/remote.js:352:123>
      at o.isAbove  <node_modules/intern/browser/remote.js:476:1019>
      at Anonymous function  <tests/unit/util.ts:159:4>
      at Anonymous function  <node_modules/intern/browser/remote.js:104:780>
      at Anonymous function  <src/util.ts:64:2>
      at run  <tests/unit/util.ts:175:4>
```

Package Version: master

edhager:
The setTimeout function on Edge doesn't seem to guarantee the delay.  Load the page below in Edge and you will see the timer fire before the requested delay.  This breaks the utils#throttle function.

```javascript
<!DOCTYPE html>
<html>
<body>
<script>
	var counter = 0;
	var start = Date.now();
	console.log('Start ', start);
	var fn = function () {
		var end = Date.now();
		var diff = end - start;
		console.log('End ', end, diff, (diff < 25) ? '<<<<<<<<<<': '');
		counter++;
		if (counter < 15) {
			start = Date.now();
			setTimeout(fn, 25);
		}
	};
	setTimeout(fn, 25);
</script>
</body>
</html>
```

PR #360, opened 2017-10-30
Guarantee util.throttle delay

edhager:
setTimeout in Edge and IE 11 do not guarantee the delay value.  In my testing, if the delay is greater than 200ms, then setTimeout is reliable but with a smaller delay, the callback function can fire early roughly 50% of the time.

I created `util.guaranteeMinimumTimeout()` which makes sure the desired amount of time has passed.

Fixes: #357

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -107,7 +107,9 @@ export function guaranteeMinimumTimeout(callback: (...args: any[]) => void, dela
 		} else {
 			// Cast setTimeout return value to fix TypeScript parsing bug.  Without it,
 			// it thinks we are using the Node version of setTimeout.
-			timerId = <any> setTimeout(timeoutHandler, delta);
+			// Revisit this with the next TypeScript update.
+			// Set another timer for the mount of time that we came up short.
+			timerId = <any> setTimeout(timeoutHandler, delay - delta);
 		}
 	}
 	timerId = setTimeout(timeoutHandler, delay);
--- a/src/util.ts
+++ b/src/util.ts
@@ -29,15 +29,15 @@ export function createTimer(callback: (...args: any[]) => void, delay?: number):
 export function debounce<T extends (this: any, ...args: any[]) => void>(callback: T, delay: number): T {
 	// node.d.ts clobbers setTimeout/clearTimeout with versions that return/receive NodeJS.Timer,
 	// but browsers return/receive a number
-	let timer: any;
+	let timer: Handle | null;
 
 	return <T> function () {
-		timer && clearTimeout(timer);
+		timer && timer.destroy();
 
 		let context = this;
 		let args: IArguments | null = arguments;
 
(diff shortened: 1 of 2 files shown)


tests/unit/util.ts
```diff
@@ -24,233 +24,278 @@ import { Handle } from '@dojo/interfaces/core';
...
-		return {
-			afterEach() {
-				timer && timer.destroy();
-				timer = null;
+	tests: {
+		createTimer: (function () {
+			let timer: Handle | null;
+
+			return {
+				afterEach() {
```
bryanforbes: Is this (and the ones below) needed with the suite-level `afterEach()`?

src/util.ts
```diff
@@ -89,9 +89,30 @@ export function throttleAfter<T extends (this: any, ...args: any[]) => void>(cal
...
+export function guaranteeMinimumTimeout(callback: (...args: any[]) => void, delay?: number): Handle {
+	const startTime = Date.now();
+	let timerId: number | null;
+
+	function timeoutHandler() {
+		const delta = Date.now() - startTime;
+		if (delay == null || delta >= delay) {
+			callback();
+		} else {
+			timerId = setTimeout(timeoutHandler);
```
bryanforbes: Should this pass `delay - delta` as the second argument to `setTimeout()`?

bryanforbes (dismissed):

src/util.ts
edhager: Yes, I goofed that when I changed it to make delay optional.  I'll fix it.

tests/unit/util.ts
edhager: I fixed it.

edhager:
The test failure "Attempt to detect unregistered has feature "web-worker-xhr-upload" is addressed in another ticket/PR: PR #361

kitsonk (approved):

src/util.ts
```diff
@@ -89,9 +89,32 @@ export function throttleAfter<T extends (this: any, ...args: any[]) => void>(cal
...
+	let timerId: number | null;
+
+	function timeoutHandler() {
+		const delta = Date.now() - startTime;
+		if (delay == null || delta >= delay) {
+			callback();
+		} else {
+			// Cast setTimeout return value to fix TypeScript parsing bug.  Without it,
+			// it thinks we are using the Node version of setTimeout.
+			timerId = <any> setTimeout(timeoutHandler, delta);
```
maier49: Should we be timing out for `delay - delta`? What we're saying here is that we want to wait as long as we've waited right? That seems arbitrary

edhager: Fixing now...

rorticus (approved):
Approved after @maier49 's comment is addressed!

maier49 (approved):

PR #361, opened 2017-10-31
Fix "random" XHR test fail.  

### LINKED ISSUE #358: Xhr unit tests intermittently fail on IE11
**Bug**

When running the unit tests on IE11, the following error is intermittently raised:

```
× internet explorer 11 on WINDOWS - request/providers/xhr - Web Workers - from blob (0.001s)
    TypeError: Attempt to detect unregistered has feature "web-worker-xhr-upload"
      at has  <node_modules/@dojo/has/has.ts:239:2>
      at from blob  <tests/unit/request/xhr.ts:567:4>
      at Anonymous function  <node_modules/intern/browser/remote.js:56:10887>
      at Anonymous function  <node_modules/intern/browser/remote.js:6:4455>
      at i  <node_modules/intern/browser/remote.js:56:27528>
      at Anonymous function  <node_modules/intern/browser/remote.js:46:5856>
      at Anonymous function  <node_modules/intern/browser/remote.js:46:5622>
      at r  <node_modules/intern/browser/remote.js:56:27750>
      at Anonymous function  <node_modules/intern/browser/remote.js:56:28694>
      at r  <node_modules/intern/browser/remote.js:56:27750>
```

This is because of the has test added to keep the test from failing because of blobs passed to web workers failing in IE11

Package Version: master

edhager:
A security error occurs in IE11 on Windows 8.1 (and probably older) in the web-worker-xhr-upload `has` test causing the `has` test to be removed from the cache.  This fix makes the `has` test return false when the security error occurs.

This test failure appeared to be random because we were asking for IE11 on "WINDOWS" in the intern configuration.  I added more specific operating system requests to intern.json.

Searching in Google revealed that the security error was occurring in Windows 10 but a fix in Windows 10 was released.

Fixes #358

### DIFF
--- a/intern.json
+++ b/intern.json
@@ -55,7 +55,8 @@
 			"tunnel": "browserstack",
 
 			"environments+": [
-				{ "browserName": "internet explorer", "version": "11" },
+				{ "browserName": "internet explorer", "version": "11", "os": "WINDOWS", "os_version": "8.1"},
+				{ "browserName": "internet explorer", "version": "11", "os": "WINDOWS", "os_version": "10"},
 				{ "browserName": "edge" },
 				{ "browserName": "firefox", "platform": "WINDOWS" },
 				{ "browserName": "chrome", "platform": "WINDOWS" }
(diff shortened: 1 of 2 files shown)


edhager:
The "utility functions - throttle - throttles callback" test will fail.  That is fixed in another PR.

rorticus (approved):
Nice catch! That must have been fun to track down 😄

bryanforbes (approved):

PR #362, opened 2017-11-03
Fix the build

maier49:
**Type:** bug

* [ ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**
Timeouts are resolving prematurely in edge on Windows 10 on BrowserStack. This works around that.


maier49:
This is addressed by #360

PR #363, opened 2017-11-04
Remove destructuring from generateRequestUrl signature

nicknisi:
**Description:**

The `generateRequestUrl` function is correct, but its usage of destructuring with a default argument is throwing an error in the way that ts-loader reads the code. Issue TypeStrong/ts-loader#442 describes the issue, but it hasn't really had much progress, and it's easier to just change this.

### DIFF
--- a/src/request/util.ts
+++ b/src/request/util.ts
@@ -8,10 +8,9 @@ import { forOf } from '@dojo/shim/iterator';
  * @param url The base URL.
  * @param options The RequestOptions used to generate the query string or cacheBust.
  */
-export function generateRequestUrl(url: string,
-		{ query, cacheBust }: RequestOptions = {}): string {
-	query = new UrlSearchParams(query).toString();
-	if (cacheBust) {
+export function generateRequestUrl(url: string, options: RequestOptions = {}): string {
+	let query = new UrlSearchParams(options.query).toString();
+	if (options.cacheBust) {
 		const bustString = String(Date.now());
 		query += query ? `&${bustString}` : bustString;
 	}


src/request/util.ts
```diff
@@ -8,10 +8,9 @@ import { forOf } from '@dojo/shim/iterator';
  * @param url The base URL.
  * @param options The RequestOptions used to generate the query string or cacheBust.
  */
-export function generateRequestUrl(url: string,
-		{ query, cacheBust }: RequestOptions = {}): string {
-	query = new UrlSearchParams(query).toString();
-	if (cacheBust) {
+export function generateRequestUrl(url: string, options: RequestOptions = {}): string {
+	let query = new UrlSearchParams(options.query).toString();
```
kitsonk, line 12: Why not just move the destructering down?

```ts
export function generateRequestUrl(url: string, options: RequestOptions = {}): string {
    let { query, cacheBust } = options;
```

kitsonk (commented):
Also the CI is failing due to some sort of TypeScript error.  Not entirely sure why but we should find out what is going on.

src/request/util.ts
nicknisi: I changed it for a couple reasons:

- I thought it'd be weird to set `query` and then set it again on the line after it.
```ts
	let { query, cacheBust } = options;
	query = new UrlSearchParams(query).toString();
```
- I didn't want `cacheBust` to be a `let` variable since it's not modified.

nicknisi:
It looks like the errors are related to errors in functional tests, and not to changes in this PR.
```
Running "ts:dev" (ts) task
Compiling...
Using tsc v2.4.2
tests/functional/text/textPlugin.ts(25,30): error TS2339: Property 'text' does not exist on type '{}'.
tests/functional/text/textPlugin.ts(30,30): error TS2339: Property 'text' does not exist on type '{}'.
tests/functional/text/textPlugin.ts(35,30): error TS2339: Property 'text' does not exist on type '{}'.
tests/functional/text/textPlugin.ts(40,30): error TS2339: Property 'text' does not exist on type '{}'.
```

I'm happy to address, but I can't reproduce locally.

src/request/util.ts
kitsonk: Yeah, ok, and it is only accessed once (`cacheBust`) 👍

kitsonk:
Found the issue and fixed it in #364.  I am not sure why it wasn't reproducible on your machine though.

rorticus (approved):

PR #367, opened 2017-11-09
Typescript 2.6

bryanforbes:
**Type:** enhancement

**Description:**

Relates to dojo/meta#210


src/QueuingEvented.ts
```diff
@@ -9,49 +9,38 @@ import Evented, { isGlobMatch, EventObject } from './Evented';
...
-			});
-
-			return handle;
-		};
 	}
+}
 
-	emit<E extends EventObject>(event: E): void {
-		super.emit(event);
+(function (proto) {
```
agubler: How come this is needed?

src/Evented.ts
```diff
@@ -170,26 +109,17 @@ export class Evented extends Destroyable implements Evented {
...
-			else {
-				return aspectOn(this.listenersMap, type, resolveListener(listeners));
-			}
-		}
-		else if (args.length === 1) {
-			const [ listenerMapArg ] = <[EventedListenersMap<any>]> args;
-			const handles = Object.keys(listenerMapArg).map((type) => this.on(type, listenerMapArg[type]));
+	on<K extends keyof M>(type: K, listener: EventedCallbackOrArray<K, M[K]>): Handle;
+	on(type: T, listener: EventedCallbackOrArray<T, O>): Handle;
+	on(type: any, listener: EventedCallbackOrArray<any, any>): Handle {
```
agubler, line 116: Assume this can't be `T | K`?

src/Evented.ts
```diff
@@ -117,36 +44,48 @@ export function isGlobMatch(globString: string | symbol, targetString: string |
...
+export class Evented<M extends {} = {}, T = EventType, O extends EventObject<T> = EventObject<T>> extends Destroyable {
 
-	/**
-	 * map of listeners keyed by event type
-	 */
-	protected listenersMap: Map<string, EventedCallback<EventObject>> = new Map<string, EventedCallback<EventObject>>();
+	// The following member is purely so TypeScript remembers the type of `M` when extending so
+	// that the utilities in `on.ts` will work
+	// tslint:disable-next-line
+	protected __typeMap__?: M;
```
agubler: This seems is a bit hacky :(

src/Evented.ts
```diff
@@ -117,36 +44,48 @@ export function isGlobMatch(globString: string | symbol, targetString: string |
...
 	 *
 	 * @param event the event to emit
 	 */
-	emit<E extends EventObject>(event: E): void {
+	emit<K extends keyof M>(event: M[K]): void;
+	emit(event: O): void;
+	emit(event: any): void {
 		this.listenersMap.forEach((method, type) => {
-			if (isGlobMatch(type, event.type)) {
+			if (isGlobMatch(<any> type, event.type)) {
```
agubler: Why do we need `<any>` here?

src/QueuingEvented.ts
```diff
@@ -9,49 +9,38 @@ import Evented, { isGlobMatch, EventObject } from './Evented';
...
+		// @ts-ignore
 		this.listenersMap.forEach((method, type) => {
 			if (isGlobMatch(type, event.type)) {
 				hasMatch = true;
 			}
 		});
 
 		if (!hasMatch) {
+			// @ts-ignore
 			let queue = this._queue.get(event.type);
```
agubler: What are we ignoring here?

src/async/timing.ts
```diff
@@ -30,10 +31,13 @@ export interface Identity<T> {
  */
 export function timeout<T>(milliseconds: number, reason: Error): Identity<T> {
 	const start = Date.now();
-	return function (value: T): Promise<T> {
+	return function (value?: IdentityValue<T>): Promise<T> {
 		if (Date.now() - milliseconds > start) {
 			return Promise.reject<T>(reason);
 		}
+		if (typeof value === 'function') {
```
agubler, line 38: Is this new functionality?

agubler (commented):
A few questions

agubler:
Would be great if @kitsonk could also review

src/async/timing.ts
agubler: Doesn't look like it's covered in tests

src/Destroyable.ts
```diff
@@ -34,12 +35,13 @@ export class Destroyable {
 	 * @param {Handle} handle The handle to add for the instance
 	 * @returns {Handle} a handle for the handle, removes the handle for the instance and calls destroy
 	 */
-	own(handle: Handle): Handle {
-		const { handles } = this;
-		handles.push(handle);
+	own(handles: Handle | Handle[]): Handle {
+		const handle = Array.isArray(handles) ? createCompositeHandle(...handles) : handles;
```
agubler, line 39: We are missing a test branch here

src/QueuingEvented.ts
kitsonk: yeah, we really should comment why we are deciding to ignore something and would much prefer casting (and a comment)...  I have yet to see a valid use case yet for `// @ts-ignore` in our code base.

src/Destroyable.ts
```diff
@@ -34,12 +35,13 @@ export class Destroyable {
...
+		const { handles: _handles } = this;
```
kitsonk, line 40: Why the rename?

src/Evented.ts
kitsonk: Yeah, feels _dirty_ to me too... 😢

src/Evented.ts
kitsonk: Yes, lets try to comment whenever we have to cast as `any` (as well as prefer the `type as any` instead of the arrow brackets.

src/QueuingEvented.ts
kitsonk: Yes, couldn't something like this be accomplished with the mixin pattern?

tests/unit/aspect.ts
```diff
@@ -4,7 +4,7 @@ const { registerSuite } = intern.getInterface('object');
 const { assert } = intern.getPlugin('chai');
 import * as sinon from 'sinon';
 import * as aspect from '../../src/aspect';
-import { Handle } from '@dojo/interfaces/core';
+import { Handle } from '../../src/interfaces';
```
kitsonk: We added support for symbols, but no tests for symbols.

src/Evented.ts
bryanforbes: Because `type` is a generic that can be anything and `isGlobMatch()` requires `string | symbol`. Also, casting `type` to `string | symbol` doesn't work either.

src/Evented.ts
bryanforbes: Correct. `K` and `T` need to be specifically associated with the correct `EventedCallbackOrArray`. Using a union type means that `T` could be associated with `EventedCallbackOrArray<K, M[K]>` which we don't want.

src/QueuingEvented.ts
bryanforbes: I'll look into the mixin pattern

bryanforbes: I think this is left-over from when the signature of `emit` and `on` were much more complex than they are now and I was trying to avoid repeating the signatures in `QueingEvented`. I've fixed these up locally to be methods on `QueingEvented`.

src/QueuingEvented.ts
bryanforbes: These were because `_queue` and `listenersMap` weren't accessible outside of the class declaration of `QueuingEvented`. Since I've moved the functions into the class definition, these comments are no longer needed.

src/async/timing.ts
bryanforbes: It's not new functionality (`value` was `T | (() => T | Thenable<T>)` before my changes), but it wasn't handled properly before. You're right that we need to test it.

src/Destroyable.ts
bryanforbes: Locally, `handles` is `Handle | Handle[]` and `this.handles` is `Handle[]`. Assigning works fine initially, but when it gets into the `destroy()` function returned, the compiler reverts to thinking that `handles` is `Handle | Handle[]` and requires a cast. I thought a rename was cleaner than a cast.

src/async/timing.ts
bryanforbes: Tests have been added

src/Destroyable.ts
bryanforbes: Tests added

src/Evented.ts
kitsonk: > let's try to comment whenever we have to cast as `any`

As in, add a comment to code for future generations, please.

src/Destroyable.ts
kitsonk: 👍

src/Evented.ts
bryanforbes: Yeah, it's dirty, but I'm not sure if there's any other way. TypeScript isn't able to remember the type map for use with the stand-alone `on()` function without it.

bryanforbes: I've filed [an issue](https://github.com/Microsoft/TypeScript/issues/20348) with a reduced test case outlining this problem. It feels like a bug.

PR #368, opened 2017-11-13
Using importHelpers instead of noEmitHelpers

rorticus:
Using `importHelpers` instead of `noEmitHelpers`. Helpers will be provided by `tslib` but augmented by `@dojo/shim/main`.

### DIFF
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -23,7 +23,7 @@ function shouldDeepCopyObject(value: any): value is Object {
 function copyArray<T>(array: T[], inherited: boolean): T[] {
 	return array.map(function (item: T): T {
 		if (Array.isArray(item)) {
-			return <any>copyArray(<any> item, inherited);
+			return <any> copyArray(<any> item, inherited);
 		}
 
 		return !shouldDeepCopyObject(item) ?
@@ -31,7 +31,7 @@ function copyArray<T>(array: T[], inherited: boolean): T[] {
 			_mixin({
 				deep: true,
 				inherited: inherited,
-				sources: <Array<T>>[item],
+				sources: <Array<T>> [ item ],
 				target: <T> {}
 			});
 	});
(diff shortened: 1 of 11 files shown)


bryanforbes (approved):

bryanforbes (changes requested):
Can you go through and make sure that where we're using `for...of` that it's not on arrays? Or, if it's on an array, we treat them differently because of https://github.com/Microsoft/TypeScript/issues/19933 .

rorticus:
@bryanforbes I replaced the array-only usages of `for..of` with `forEach`

src/lang.ts
```diff
@@ -51,13 +52,15 @@ function _mixin<T extends {}, U extends {}>(kwArgs: MixinArgs<T, U>): T&U {
...
+	for (let i = 0; i < kwArgs.sources.length; i++) {
+		const source = kwArgs.sources[ i ];
+
 		if (source === null || source === undefined) {
 			continue;
 		}
 		for (let key in source) {
 			if (inherited || hasOwnProperty.call(source, key)) {
-				let value: any = source[key];
+				let value: any = source[ key ];
```
bryanforbes: Why are you adding spaces around `key`?

rorticus:
@bryanforbes Webstorm was overzealous. I've put back (most) of the style changes. I left ones alone that were fixing previous style errors because of reasons.

src/load/util.ts
```diff
@@ -36,7 +36,16 @@ export function isPlugin(value: any): value is LoadPlugin<any> {
...
+	if (isArrayLike(modules)) {
+		let processedModules: any[] = [];
+
+		for (let i = 0; i < modules.length; i++) {
+			const module = modules[i];
+			processedModules.push((module.__esModule && module.default) ? module.default : module);
+		}
+
+		return processedModules;
+	} else if (isIterable(modules)) {
```
bryanforbes, line 48: Should the `else if` be on a new line?

bryanforbes (changes requested):

src/load/util.ts
rorticus: it is now!

bryanforbes (approved):

PR #373, opened 2017-12-14
Move compare from dojo/test-extras to dojo/core

edhager:
dojo/test-extras/compare and dojo/core/compare where nearly identical.  I copied the more recent updates to dojo/test-extras/compare into dojo/compare.  Then I updated dojo/compare to support typed array.

Resolves dojo/test-extras#75.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@dojo/core",
-  "version": "0.3.1-pre",
+  "version": "0.3.1-rc.1",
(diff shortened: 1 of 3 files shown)


dylans (changes requested):
This needs to be rebased (assuming it is still needed)

agubler:
I am not sure this is needed now, `test-extras` doesn't use compare any more - closing, we can reopen if the consensus is that these changes are still required.

PR #377, opened 2018-01-10
Make tslib a dependency instead of a dev-dependency

### LINKED ISSUE #372: tslib must be a dependency and not a devDependency
When using `importHelpers: true`, tslib must be a runtime dependency, currently it is a development dependency so importing `@dojo/core` throws an error `Error: Cannot find module 'tslib'`

Pull introducing error: #367

mgroenhoff:
**Type:** bug

* [ ] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

Resolves #372

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7877,8 +7877,7 @@
     "tslib": {
       "version": "1.8.1",
       "resolved": "https://registry.npmjs.org/tslib/-/tslib-1.8.1.tgz",
-      "integrity": "sha1-aUavLR1lGnsYY7Ux1uWvpBqkTqw=",
-      "dev": true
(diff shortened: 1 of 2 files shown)


agubler (approved):
Thanks @mgroenhoff!

matt-gadd:
should we raise a `@dojo/meta` issue for this? it should be a dependency across all of the @dojo packages.

mgroenhoff:
Good point. I did not look beyond `@dojo/core` and its dependencies but a quick look at other packages shows that at least also `@dojo/widgets` does have `"importHelpers": true` while not having any dependency on `tslib`. Maybe there are more!

_Edit:_ I found some more and went ahead and created the issue https://github.com/dojo/meta/issues/226

PR #388, opened 2018-04-05
Make test request using currently active protocol

### LINKED ISSUE #387: has.ts causes loading the main bundle to crash in Internet Explorer 11 in https
**Bug**

When using a bundled application with Internet Explorer 11 together with `https` the bundle is not loaded. IE raises an "access denied" error in the main bundle. 

Package Version: dojo RC1 (and RC2)

**Code**

can be tested with:
* create a fresh app with dojo/cli
* add the following code after the imports:

`
import request from '@dojo/core/request';
console.log(request);
`

* pack the application with `dojo build -m dist`
* access the application via https://localhost:... in IE 11

**Actual behavior:**

IE 11 will not show the rotating elements - it will show an error in the console (no SVG, ... )

jason0x43:
For the blob test, make the test request using the active protocol rather than always using http. At least IE11 will throw an exception when making an http request from an https domain.

**Type:** bug

* [x] Unit or Functional tests are included in the PR (N/A)

**Description:**

Resolves #387

### DIFF
--- a/src/has.ts
+++ b/src/has.ts
@@ -19,7 +19,7 @@ add(
 		}
 
 		const request = new global.XMLHttpRequest();
-		request.open('GET', 'http://www.google.com', true);
+		request.open('GET', global.location.protocol + '//www.google.com', true);
 		request.responseType = 'blob';
 		request.abort();
 		return request.responseType === 'blob';


src/has.ts
```diff
@@ -19,7 +19,7 @@ add(
 		}
 
 		const request = new global.XMLHttpRequest();
-		request.open('GET', 'http://www.google.com', true);
+		request.open('GET', global.location.protocol + '//www.google.com', true);
```
dasa, line 22: I think it'd be better to hard-code `https` since `location.protocol` is `blob:` in a Worker created from a Blob.

szwoelf: I did my tests also with "https" - to be exact with "https://www.google.com". I'd suggest to use it exactly like in Dojo1 (and possibly add the comment that is also there.)
```
+               // From dojo1: https://github.com/dojo/dojo/blob/master/request/xhr.js#L37
+               // The URL used here does not have to be reachable as the XHR's `send` method is never called.
+               // It does need to be parsable/resolvable in all cases, so it should be an absolute URL
+               // and it must be an https based URL for Internet Explorer 11 (when https is used).
+               // XMLHttpRequest within a Worker created from a Blob does not support relative URL paths.
+               request.open('GET', 'https://dojotoolkit.org', true);

```

jason0x43: > I think it'd be better to hard-code https since location.protocol is blob: in a Worker created from a Blob.

Not that I'm really opposed, but why? As the comment in the dojo 1 code points out, it needs to be a parseable URL, and for IE it also must be compatible with the current location (which means https when coming from https in IE11). `global.location.protocol` works because you're just using the current protocol. Https also works, it just seems kind of arbitrary.

dasa: I'm assuming that `blob://www.google.com` would give an error.

jason0x43: It works; I even tested it in an HTTPS app in IE11. (I actually started out with a protocol relative URL, but Web Workers in FF don't like that.) The test URL doesn't have to be an active endpoint, just something the browser _could_ make a request to.

dasa: 👍

szwoelf: Great! Thanks.

dylans (approved):

rorticus (approved):

PR #391, opened 2018-04-27
Move createHandle’s destroy from context to a flag

pottedmeat:
**Type:** bug

* [ ] All code has been formatted with [`prettier`](https://prettier.io/) as per the [readme code style guidelines](./../#code-style)

**Description:**

If the destroy function in the response object from `createHandle` was called in a different context than the original, `destroy` would be assigned to the call object and the `destructor` function would also be called in that context. By using a flag, we can guarantee the destructor is only called once regardless of associated context.

### DIFF
--- a/src/lang.ts
+++ b/src/lang.ts
@@ -364,7 +364,7 @@ export function partial(targetFunction: (...args: any[]) => any, ...suppliedArgs
  * @return The handle object
  */
 export function createHandle(destructor: () => void): Handle {
-	var called = false;
+	let called = false;
 	return {
 		destroy: function(this: Handle) {
 			if (!called) {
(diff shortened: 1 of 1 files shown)


src/lang.ts
```diff
@@ -364,10 +364,13 @@ export function partial(targetFunction: (...args: any[]) => any, ...suppliedArgs
  * @return The handle object
  */
 export function createHandle(destructor: () => void): Handle {
+	var called = false;
```
bryanforbes: Should this be `let`?

pottedmeat: Ugh, I've been working with dojo 1

pottedmeat: I even had the red squiggly

bryanforbes (approved):

dylans (approved):

PR #394, opened 2018-05-02
Fix readme punctuation

dylans:
README punctuation fix for consistency.

[ci skip]

### DIFF
--- a/docs/on.md
+++ b/docs/on.md
@@ -33,7 +33,7 @@ on(button, 'click', function (event) {
 });
 
 ```
-## .once.
+## 'once'
 
 Attach an event that can only be called once to a target.
 


docs/on.md
```diff
@@ -33,7 +33,7 @@ on(button, 'click', function (event) {
 });
 
 ```
-## .once.
+## 'once'
```
nicknisi, line 36: Should these be backticks?

dylans: Maybe, but they are not for the other items in this readme...

nicknisi: Ah, I see that was an oversight in my previous commit. They should all be backticks in this file.

PR #399, opened 2018-05-31
add support for signal request option

### LINKED ISSUE #390: Support the abortable fetch API
**Enhancement**

Add support for a `signal: AbortSignal` property to `RequestOptions`

Package Version: 2.x

**Code**

This would allow `core/request` to support the [Abortable fetch](https://developers.google.com/web/updates/2017/09/abortable-fetch) pattern.

There'd need to be a shim for `AbortController` too.

nicknisi:
**Type:** feature

**Description:**

**Note:** This PR is dependent on dojo/shim#143.

Resolves #390

add support for a `signal: AbortSignal` property in the request options. When provided, the request can be aborted by calling the `AbortController#abort` method, similar to the fetch API. When a request is aborted, the `Promise` is rejected with an `AbortError`, if possible.

### DIFF
--- a/src/request/providers/node.ts
+++ b/src/request/providers/node.ts
@@ -654,6 +654,15 @@ export default function node(url: string, options: NodeRequestOptions = {}): Upl
 					timeoutReject && timeoutReject(new TimeoutError('The request timed out'));
 				}, options.timeout);
 			}
+
+			if (options.signal) {
+				options.signal.addEventListener('abort', () => {
+					const abortError = new Error('Aborted');
+					abortError.name = 'AbortError';
+					reject(abortError);
+				});
+			}
+
 		},
 		() => {
 			request.abort();
@@ -670,10 +679,6 @@ export default function node(url: string, options: NodeRequestOptions = {}): Upl
 		error.message = '[' + requestOptions.method + ' ' + sanitizedUrl + '] ' + error.message;
 		throw error;
 	});
-
-	if (options.signal) {
-		options.signal.addEventListener('abort', () => requestTask.cancel());
-	}
 
 	requestTask.upload = new Observable<number>((observer) => uploadObserverPool.add(observer));
 
--- a/src/request/providers/node.ts
+++ b/src/request/providers/node.ts
@@ -671,6 +671,10 @@ export default function node(url: string, options: NodeRequestOptions = {}): Upl
 		throw error;
 	});
 
+	if (options.signal) {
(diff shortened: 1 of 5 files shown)


rorticus (approved):

src/request/providers/xhr.ts
```diff
@@ -205,6 +205,19 @@ export default function xhr(url: string, options: XhrRequestOptions = {}): Uploa
...
+		if (options.signal) {
+			options.signal.addEventListener('abort', () => {
+				let abortError: Error;
+				try {
+					abortError = new DOMException('Aborted', 'AbortError');
+				} catch {
+					abortError = new Error('Aborted');
+					abortError.name = 'AbortError';
+				}
+				reject(new DOMException('Aborted', 'AbortError'));
```
dasa: Looks like you should use `abortError` here.

nicknisi: Nice catch! Thanks!

src/request/providers/xhr.ts
```diff
@@ -205,6 +205,19 @@ export default function xhr(url: string, options: XhrRequestOptions = {}): Uploa
 	const task = <UploadObservableTask<XhrResponse>>new Task<XhrResponse>((resolve, reject) => {
 		timeoutReject = reject;
 
+		if (options.signal) {
+			options.signal.addEventListener('abort', () => {
```
ycabon, line 209: when the request is fulfilled, the event listener should be removed.

src/request/providers/node.ts
```diff
@@ -654,6 +654,15 @@ export default function node(url: string, options: NodeRequestOptions = {}): Upl
 					timeoutReject && timeoutReject(new TimeoutError('The request timed out'));
 				}, options.timeout);
 			}
+
+			if (options.signal) {
+				options.signal.addEventListener('abort', () => {
```
ycabon, line 659: when the request is fulfilled, the event listener should be removed.

src/request/providers/xhr.ts
nicknisi: @ycabon I'm not sure there's a good place in the code to remove this event listener. Also, I'm not sure it's necessary as subsequent `abort` events will have no effect on a fulfilled Task.

ycabon: My concern was more for cases where the signal is reused for multiple calls, would the Task be kept in memory until the signal and controller are GCed.

PR #408, opened 2018-06-12
Make a copy of the Evented listeners in case one is removed during an emit

### LINKED ISSUE #407: Evented listener destroying its own handle won't run all listeners
**Bug** 

Package Version: 2.0.0

**Code**

```typescript
const e = new Evented();
e.on('foo', () => console.log(1));
const h = e.on('foo', () => { console.log(2); h.destroy(); });
e.on('foo', () => console.log(3));
e.emit({ type: 'foo' });
```

**Expected behavior:**

Logs 1, 2, and 3

**Actual behavior:**

Logs 1 and 2, but not 3

bryanforbes:
**Type:** bug

**Description:**

Resolves #407

### DIFF
--- a/src/Evented.ts
+++ b/src/Evented.ts
@@ -77,7 +77,7 @@ export class Evented<
 	emit(event: any): void {
 		this.listenersMap.forEach((methods, type) => {
 			if (isGlobMatch(type as any, event.type)) {
-				methods.forEach((method) => {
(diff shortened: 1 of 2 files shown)


agubler (approved):
I can't think of a reason why cloning the array of methods would cause any issues.

rorticus (approved):

